PR #1, opened 2024-03-06
[DPE-3517] Kyuubi Charm Implementation

theoctober19th:
This PR creates the initial draft implementatin of Apache Kyuubi charm.
The tests are yet to be added and will be added via a different PR.

DPE-3517, DPE-3751


.env
```diff
@@ -0,0 +1 @@
+PYTHONPATH=lib:src
```
welpaolo, line 1: Do we really need this file?

README.md
```diff
@@ -0,0 +1,11 @@
+# kyuubi
```
welpaolo, line 1: Capital K

welpaolo: Maybe a good name should be Kyuubi-k8s-operator. What do you think @deusebio ?

README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This is a Kubernetes charm for [Apache Kyuubi](https://kyuubi.apache.org/).
+
+Charmhub package name: kyuubi
```
welpaolo, line 5: same here

requirements.txt
```diff
@@ -0,0 +1,4 @@
+ops ~= 2.5
```
welpaolo, line 1: I will honestly start directly by using Poetry since all our charms use it,

src/charm.py
```diff
@@ -0,0 +1,165 @@
...
+        self.update_service()
+
+    def _update_event(self, _):
+        """Handle the update event hook."""
+        self.unit.status = self.get_status()
+
+    def _update_spark_configs(self):
+        """Update Spark properties in the spark-defaults file inside the charm container."""
+        s3_info = self.s3_connection_info
+        namespace = self.config["namespace"]
```
welpaolo, line 73: You can use constants/literals for the configuration options

src/charm.py
```diff
@@ -0,0 +1,165 @@
...
+        status = self.get_status()
+        self.unit.status = status
+
+        if status is not Status.ACTIVE.value:
+            self.logger.info(f"Cannot start service because of status {status}")
+            self.workload.stop()
+            return False
+
+        # Dynamically update the Spark properties
+        self._update_spark_configs()
```
welpaolo, line 121: nitpick: maybe it is not need to update the spark configs on any call of the `update_service` function

src/charm.py
```diff
@@ -0,0 +1,165 @@
...
+            return Status.WAITING_PEBBLE.value
+
+        s3_info = self.s3_connection_info
+        if not s3_info:
+            return Status.MISSING_S3_RELATION.value
+
+        if not s3_info.verify():
+            return Status.INVALID_CREDENTIALS.value
+
+        namespace = self.config["namespace"]
```
welpaolo, line 95: Constant/literals

src/config.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python3
+
+"""Spark History Server configuration."""
```
welpaolo, line 3: Update the description of the file from Spark History Server to Kyuubi charm

src/k8s_utils.py
```diff
@@ -0,0 +1,23 @@
...
+def is_valid_namespace(namespace: str) -> bool:
+    """Return whether given namespace exists in K8s cluster."""
+    try:
+        Client().get(Namespace, name=namespace)
+    except ApiError:
+        return False
+    return True
+
+
+def is_valid_service_account(namespace: str, service_account: str) -> bool:
```
welpaolo, line 17: Here I will also check if there are enough permission to run charmed spark jobs, you can re-use some of the work we did in the spark8t library.

src/models.py
```diff
@@ -0,0 +1,27 @@
...
+    group: str = "spark"
+
+
+class Status(Enum):
+    """Class bundling all statuses that the charm may fall into."""
+
+    WAITING_PEBBLE = MaintenanceStatus("Waiting for Pebble")
+    MISSING_S3_RELATION = BlockedStatus("Missing S3 relation")
+    INVALID_CREDENTIALS = BlockedStatus("Invalid S3 credentials")
+    MISSING_INGRESS_RELATION = BlockedStatus("Missing INGRESS relation")
```
welpaolo, line 23: I believe this option is not needed at the moment

src/s3.py
```diff
@@ -0,0 +1,40 @@
+"""S3 connection utility classes and methods."""
```
welpaolo, line 1: OUT OF SCOPE: for future reference please have a look at the new implementation of this class in the case of self-signed certificate

tests/integration/test_charm.py
```diff
@@ -0,0 +1,38 @@
...
+from pytest_operator.plugin import OpsTest
+
+logger = logging.getLogger(__name__)
+
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+APP_NAME = METADATA["name"]
+
+
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy(ops_test: OpsTest):
```
welpaolo, line 16: This test will never  pass so please remove it from the PR

tox.ini
```diff
@@ -0,0 +1,84 @@
...
+
+[testenv:static]
+description = Run static type checks
+deps =
+    pyright
+    -r {tox_root}/requirements.txt
+commands =
+    pyright {posargs}
+
+[testenv:integration]
```
welpaolo, line 71: Remove or comment also this entry in the tox file

tests/unit/test_charm.py
```diff
@@ -0,0 +1,63 @@
+import unittest
+
+import ops
+import ops.testing
+from charm import KyuubiCharm
+
+
+class TestCharm(unittest.TestCase):
```
welpaolo, line 8: Please add some minimal unit tests

welpaolo (commented):
Looks a good starting point, I left some more specific comments in the code. 

I leave below some general comments: 
- Copyrights are missing everywhere, please add them
- I will use Poetry to have a better handling of python dependencies
- Add missing docstrings in the different Python files
- Please add some basic unit tests
- The Github CI/CD is completely missing please add it

theoctober19th:
Thanks @welpaolo. I'll move the PR back to draft, address these comments and then mark it as ready to review again.

.env
theoctober19th: This file is not needed in prod, and is only for VSCode to recognize the lib directory. I'll remove this from the upstream source and make git ignore this file.

README.md
theoctober19th: Taking reference from the title in history server repo, shall I rename to "Charmed Kyuubi Operator"?

README.md
theoctober19th: I'm more inclined to use `kyuubi-k8s` as the name, following the similar pattern we do for `spark-history-server-k8s`. I'll update this accordingly.

tests/unit/test_charm.py
theoctober19th: Added.

src/k8s_utils.py
theoctober19th: I tried checking the permissions with `kubectl auth can-i --as sytem:serviceaccount:foo:bar`, however this fails inside the context of the charm unit.

The charm unit pod assumes the role of `system:serviceaccount:model_name:app_name` which is created by default by Juju, and this service account is not able to impersonate other service accounts to check RBAC with `--as` option. And as this check will always run inside the context of unit pod, I was not able to find a reliable way to check the actual permissions.

In Spark8t, we used these checks during integration tests which run in the context of edge machine (which has role of `admin` user account).

Let me know if you think there's workaround around this, I'm open for discussion.

src/s3.py
theoctober19th: Thanks.

README.md
theoctober19th: Title changed to "Kyuubi K8s Operator" for now.

src/charm.py
```diff
@@ -0,0 +1,168 @@
...
+        # Start the workload
+        self.workload.start()
+        return True
+
+    def _on_kyuubi_pebble_ready(self, event: ops.PebbleReadyEvent):
+        """Define and start a workload using the Pebble API."""
+        logger.info("Kyuubi pebble service is ready.")
+        self.update_service()
+
+    def _on_get_jdbc_endpoint(self, event: ActionEvent):
```
deusebio, line 135: **out-of-scope** this will probably be replaced by some relations (e.g. data-integrator)

src/k8s_utils.py
```diff
@@ -0,0 +1,28 @@
...
+# See LICENSE file for licensing details.
+
+"""Utility methods related to Kubernetes resources."""
+
+from lightkube import Client
+from lightkube.core.exceptions import ApiError
+from lightkube.resources.core_v1 import Namespace, ServiceAccount
+
+
+def is_valid_namespace(namespace: str) -> bool:
```
deusebio, line 13: **out-of-scope, refactoring** these functionalities seems quite good for a "k8s manager"

src/workload.py
```diff
@@ -0,0 +1,77 @@
...
+class KyuubiServer(WithLogging):
+    """The abstraction of Kyuubi workload container."""
+
+    def __init__(self, container: Container, user: User = User()):
+        self.container = container
+        self.user = User
+
+    def get_spark_configuration_file(self, mode: IOMode) -> ContainerFile:
+        """Return the configuration file for Spark History server."""
+        daemon_user = User(name="_daemon_", group="_daemon_")
```
deusebio, line 26: **todo** why not using `self.user`? I believe you are now using the `self.user`, you could just replace the fact that we use "spark" with using `_daemon_`?

**out-of-scope, thought** Maybe we could consider to run the pebble services (not pebble) as the spark user. But let's think about this and discuss it first

deusebio (commented):
The codes look all right, although we'll need to restructure this a bit eventually to have the same pattern as the spark-history server. When doing that refactoring, please follow up on some of the **out-of-scope, refactoring** comments I left. 

Right now, the only part worth discussing (and addressing) would be to use the spark8t library to create and read/parse the configuration. This way the business logic for the config should become much simpler. 

Right now we may require the user to create the service account beforehand, but eventually (once we will have the configuration hub) we could probably just require to have a relation with the configuration hub. We should probably discuss this UX, though.

src/charm.py
theoctober19th: That is noted. The action is a temporary implementation so we get the first working implementation of the charm ready.

src/k8s_utils.py
theoctober19th: Noted.

src/workload.py
theoctober19th: Updated to use `self.user` while the `User` class has `_daemon_` as default values for `user` and `group`.

deusebio (approved):
Looks good! As discussed, we may need to address few points, but we can do this as further improvements and as part of the refactoring

.github/workflows/release.yaml
```diff
@@ -0,0 +1,28 @@
+name: Release to charmhub
```
welpaolo, line 1: Copyright

.github/workflows/sync_issue_to_jira.yaml
```diff
@@ -0,0 +1,22 @@
+# Copyright 2023 Canonical Ltd.
```
welpaolo, line 1: Update copyright date

src/charm.py
welpaolo: That's fine then

tests/unit/conftest.py
```diff
@@ -0,0 +1,80 @@
+# Copyright 2023 Canonical Limited
```
welpaolo, line 1: Copyright date

welpaolo (approved):
Really good step forward! 
Really little and minor comments

PR #4, opened 2024-03-20
[DPE-3752] Write integration tests for Kyuubi charm

theoctober19th:
DPE-3752
https://warthogs.atlassian.net/browse/DPE-3752


tests/integration/conftest.py
```diff
@@ -0,0 +1,166 @@
...
+
+logger = logging.getLogger(__name__)
+
+TEST_BUCKET_NAME = "kyuubi-test"
+TEST_NAMESPACE = "kyuubi-test"
+TEST_SERVICE_ACCOUNT = "kyuubi-test"
+TEST_POD_SPEC_FILE = "./tests/integration/setup/testpod_spec.yaml"
+
+
+class MockCharm(BaseModel):
```
deusebio, line 25: **question** Why is this called `MockCharm`? This does not seem mock (like a test charm) but rather some charm metadata for the version. I checked on Spark History Server, and indeed, we call this similar class `CharmVersion` (see [here](https://github.com/canonical/spark-history-server-k8s-operator/blob/main/tests/integration/conftest.py#L10)).

deusebio (approved):
Really solid and good work! Well done @theoctober19th !

A couple of minor nonblocking thoughts/suggestion

tests/integration/conftest.py
theoctober19th: `CharmVersion` name seemed a bit off to me because the first time I read the name I mistook the class is about a version. Maybe `MockCharm` could be replaced with `TestCharm`, just so that we refrain from using the word "mock" (which in this case it is not).

tests/integration/conftest.py
```diff
@@ -0,0 +1,166 @@
...
+def charm_versions() -> IntegrationTestsCharms:
+    return IntegrationTestsCharms(
+        s3=MockCharm(
+            **{"name": "s3-integrator", "channel": "edge", "series": "jammy", "alias": "s3"}
+        ),
+    )
+
+
+@pytest.fixture(scope="module")
+def s3_bucket_and_creds():
```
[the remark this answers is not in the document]
theoctober19th: Thank you!

.github/workflows/ci.yaml
```diff
@@ -66,56 +66,56 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    timeout-minutes: 120
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install Java
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: '8'
```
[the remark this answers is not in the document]
theoctober19th: I agree, I'll bump it to 17.

tests/integration/conftest.py
deusebio: yes, naming is a mess. Let's discuss this later to find a suitable name we all like :D

theoctober19th: This has been updated to `TestCharm`.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
theoctober19th: Bumped the java version to 17.

welpaolo (approved):
LGTM

PR #6, opened 2024-03-31
[DPE-3007] Add Support to use PostgreSQL as Hive Metastore

theoctober19th:
Add support to use PostgredSQL as Hive Metastore by integrating `kyuubi-k8s` with `postgresql-k8s`.


welpaolo (commented):
The PR looks good. I left some minor comments.
I believe that we are missing some integrations tests:
- Removal of integration with PostgreSQL and check that everything is working
- Integrating again with PostgreSQL and check the existence of the previously created tables and everything is working properly

src/charm.py
```diff
@@ -153,6 +183,22 @@ def s3_connection_info(self) -> Optional[S3ConnectionInfo]:
             bucket=raw_info.get("bucket"),
         )
 
+    @property
+    def db_connection_info(self) -> Optional[DatabaseConnectionInfo]:
```
deusebio, line 187: Following  @welpaolo comment, it is probably best to rename the property to refer to `metastore`

charmcraft.yaml
```diff
@@ -56,6 +68,9 @@ resources:
...
     upstream-source: ghcr.io/canonical/charmed-spark:3.4-22.04_edge
 
+
 requires:
   s3-credentials:
-    interface: s3
\ No newline at end of file
+    interface: s3
+  database:
+    interface: postgresql_client
```
deusebio, line 76: **todo** we should probably put a limit on the relations. I don't think we could have more than 1 metastore

src/charm.py
```diff
@@ -40,10 +50,13 @@ def __init__(self, *args):
         super().__init__(*args)
         self.workload = KyuubiServer(self.unit.get_container(KYUUBI_CONTAINER_NAME))
         self.s3_requirer = S3Requirer(self, S3_INTEGRATOR_REL)
+        self.database = DatabaseRequires(
```
deusebio, line 53: maybe we could call it `self.metastore`

tests/integration/test_charm.py
```diff
@@ -152,6 +157,103 @@ async def test_jdbc_endpoint(ops_test: OpsTest, test_pod):
...
+        cursor.execute(""" SELECT * FROM "DBS" WHERE "NAME" != 'default' """)
+        num_dbs = cursor.rowcount
+        cursor.execute(""" SELECT * FROM "TBLS" """)
+        num_tables = cursor.rowcount
+
+    connection.close()
+
+    # Assert that new database and tables have indeed been added to metastore
+    assert num_dbs != 0
+    assert num_tables != 0
```
deusebio, line 254: **todo** Can we check also that when we remove the relation, we fall back on using the derby hive metastore?

src/charm.py
```diff
@@ -53,12 +66,25 @@ def register_event_handlers(self):
...
         """Handle the `on_install` event."""
         self.unit.status = Status.WAITING_PEBBLE.value
 
+    def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
+        logger.info("Database created...")
+        self.update_service()
+
+    def _on_database_relation_removed(self, event) -> None:
+        logger.info("Database relation removed")
+        self.update_service()
```
deusebio, line 86: **todo** can we do a double check that this works as expected? both manually and with integration tests? The reason is that database removal events are somewhat curious, in that the relation still exists but it about to be torn down. What this means is that the data is still there and often in the logic we do checks on relation and/or data existence, thus not updating the service by removing the properties from the configuration files, until another `update_status` event is triggered.

This is actually the reason why in some removed events (in the new pattern, BTW at a certain point we should refactor the charm into the new pattern), we force the setting to "None" (see [here](https://github.com/canonical/spark-history-server-k8s-operator/blob/main/src/events/s3.py#L53)), to mimic that the relation was torn down (although the data is still there).

deusebio (commented):
The changes look already rather good! So I believe we should be close. There is one point that I'd like to investigate to make sure that the behaviour is as expected. 

If it is has expected, it could be that this is abstracted by the DP libs (by [this](https://github.com/canonical/kyuubi-k8s-operator/blob/postgresql-integration/lib/charms/data_platform_libs/v0/data_interfaces.py#L802) function). If so, this function seems actually quite nice and worth to be used more when creating data-models in the `core` submodule.

charmcraft.yaml
```diff
@@ -56,6 +68,9 @@ resources:
     description: OCI image for kyuubi
     upstream-source: ghcr.io/canonical/charmed-spark:3.4-22.04_edge
 
+
 requires:
   s3-credentials:
-    interface: s3
\ No newline at end of file
+    interface: s3
+  database:
```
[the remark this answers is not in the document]
theoctober19th: How about `metastore-db`, just like `auth-db` for authentication database?

tests/integration/test_charm.py
```diff
@@ -6,10 +6,13 @@
...
+import psycopg2
 import pytest
 import yaml
 from pytest_operator.plugin import OpsTest
 
+from constants import METASTORE_DATABASE_NAME
+
 logger = logging.getLogger(__name__)
 
 APP_NAME = "kyuubi-k8s"
```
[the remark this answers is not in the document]
theoctober19th: I think APP_NAME can be inferred from `charmcraft.yaml` file itself and `BUCKET_NAME` can be removed, thus eliminating the necessity for a constant file.

tests/integration/test_charm.py
```diff
@@ -152,6 +157,103 @@ async def test_jdbc_endpoint(ops_test: OpsTest, test_pod):
...
+    kyuubi_unit = ops_test.model.applications[APP_NAME].units[0]
+    action = await kyuubi_unit.run_action(
+        action_name="get-jdbc-endpoint",
+    )
+    result = await action.wait()
+
+    jdbc_endpoint = result.results.get("endpoint")
+    logger.info(f"JDBC endpoint: {jdbc_endpoint}")
+
+    pod_name = test_pod
```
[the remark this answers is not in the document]
theoctober19th: Updated.

src/charm.py
theoctober19th: Changed to `self.metastore_db`. The other for the authentication will be `self.auth_db`.

charmcraft.yaml
theoctober19th: Updated to `limit: 1`.

src/charm.py
theoctober19th: Updated to `metastore_db_connection_info`.

src/charm.py
theoctober19th: During my tests, I've always found the `hive-site.xml` is updated by the relation broken event and never encountered the edge case you mentioned. I've also added a test case to check the removal of relation in the integration tests.

theoctober19th:
> The PR looks good. I left some minor comments. I believe that we are missing some integrations tests:
> 
> * Removal of integration with PostgreSQL and check that everything is working
> * Integrating again with PostgreSQL and check the existence of the previously created tables and everything is working properly

I have added a test for 1. However, as we discussed yesterday, since we do not guarantee the persistence / removal of metadata upon the change of metastore from A to B to A back again, I have skipped 2 at the moment. WDYT @welpaolo @deusebio ?

deusebio:
That's fine by me! We can maybe just have a tech debt in the backlog for this though. The use case that Paolo is suggesting is relevant, but a nice to have at the moment

welpaolo:
Fine with me as well!

welpaolo (approved):

deusebio (approved):
All good!

PR #8, opened 2024-04-15
[DPE-3936] Add authentication support for Kyuubi charm

theoctober19th:
* Add relation with Postgresql over auth-db interface.
* Create authentication database, tables and default users upon integration with Postgres
* Delete authentication database when relation removed from Postgres
* Expose `get-password` and `set-password` actions
* Integration tests


src/database.py
```diff
@@ -21,17 +23,49 @@ class DatabaseConnectionInfo(WithLogging):
     username: str
     password: str
 
-    def verify(self) -> bool:
-        """Verify the PostgreSQL connection."""
+    def execute(self, dbname: str, query: str, vars=None) -> tuple[bool, list]:
```
deusebio, line 26: **suggestion** maybe there is a use-case for having the database name as part of the attributes of the class (together with username, password, etc). This was the code would read better I believe, since we centralize these information into a single class. Right now we pass the database name around separately, with risk of inconsistency

src/config/kyuubi.py
```diff
@@ -7,72 +7,43 @@
...
+    """Kyuubi Configurations."""
 
+    def __init__(self, db_info: Optional[DatabaseConnectionInfo]):
+        self.db_info = db_info
 
-class KyuubiServerConfig(WithLogging):
-    """Spark History Server Configuration."""
+    def _get_db_connection_url(self) -> str:
+        endpoint = self.db_info.endpoint
+        return f"jdbc:postgresql://{endpoint}/{AUTHENTICATION_DATABASE_NAME}"
```
deusebio, line 23:
```suggestion
        return f"jdbc:postgresql://{endpoint}/{self.db_info.name}"
```

src/utils/auth.py
```diff
@@ -0,0 +1,109 @@
...
+class Authentication(WithLogging):
+    """Class representing authentication."""
+
+    database: DatabaseConnectionInfo
+
+    def create_authentication_table(self) -> bool:
+        """Create authentication table in the authentication database."""
+        self.logger.info("Creating authentication table...")
+        query = f"""
+            CREATE TABLE {AUTHENTICATION_TABLE_NAME} (
```
deusebio, line 32:
```suggestion
            CREATE TABLE {self.database.table} (
```

src/utils/auth.py
```diff
@@ -0,0 +1,109 @@
...
+            password,
+            username,
+        )
+        status, _ = self.database.execute(
+            dbname=AUTHENTICATION_DATABASE_NAME, query=query, vars=vars
+        )
+        if not status:
+            raise Exception(f"Could not update password of {username}.")
+
+    def create_admin_user(self) -> bool:
```
deusebio, line 90: **suggestion** I'm not sure it is really needed to have such a method. I would believe it is clearer to have an explicit call 

```
self.create_user(DEFAULT_ADMIN_USERNAME, self.generate_password())
```

Where we create the admin. I don't think there is too much value here. But it is not a big point

src/utils/auth.py
```diff
@@ -0,0 +1,109 @@
...
+        """Create authentication table in the authentication database."""
+        self.logger.info("Creating authentication table...")
+        query = f"""
+            CREATE TABLE {AUTHENTICATION_TABLE_NAME} (
+                id SERIAL PRIMARY KEY,
+                username VARCHAR(100) UNIQUE NOT NULL,
+                passwd VARCHAR(255) NOT NULL
+            );
+        """
+        status, _ = self.database.execute(AUTHENTICATION_DATABASE_NAME, query)
```
deusebio, line 38:
```suggestion
        status, _ = self.database.execute(self.database.name, query)
```

src/database.py
deusebio: Actually maybe we can have a dataclass that encapsulate all information: database name, table name, endpoint, etc, that can be passed around. This makes the objects a bit transparent, with a single dependency, instead of overusing constants (that are subtle, not transparent depenencies - less functional)

src/utils/auth.py
```diff
@@ -0,0 +1,109 @@
...
+# Copyright 2024 Canonical Limited
+# See LICENSE file for licensing details.
+
+"""S3 connection utility classes and methods."""
+
+import secrets
+import string
+from dataclasses import dataclass
+
+from constants import (
```
deusebio, line 12: **suggestion** I would try to not use these imports here

deusebio (commented):
The changes look ok, but we should really do a refactoring of the codebase. When I see the code, I can identify the different blocks (core, managers and events), but things are still a bit mixed and without a clear organization in the code just yet. 

As usual (:D ) the only issue I believe it would be worth addressing here is to inherit from the spark8t service account configuration written to k8s. I believe I raised this last time as well, but I don't remember why we did not follow up and the rationale to "replicate" the business logic around "generating" the spark properties.

src/utils/utils.py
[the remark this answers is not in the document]
theoctober19th: My bad, this file is to be removed since all of its content has already been moved to utils.* modules. Thanks for pointing it out.

src/utils/auth.py
```diff
@@ -0,0 +1,109 @@
...
+    AUTHENTICATION_TABLE_NAME,
+    DEFAULT_ADMIN_USERNAME,
+    POSTGRESQL_DEFAULT_DATABASE,
+)
+from database import DatabaseConnectionInfo
+from utils.utils import WithLogging
+
+
+@dataclass
+class Authentication(WithLogging):
```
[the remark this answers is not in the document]
theoctober19th: Noted.

src/database.py
theoctober19th: The reason I have not included the database name as an attribute of the dataclass is because the same class might need to connect to multiple databases and not a single database. For example, when creating the authentication database fixture, the charm needs to connect to `auth_db` database, however, when removing the fixture, it should connect to some other database (in this case `postgres`) because one cannot remove the database it is connected to.

In case of `auth`, it makes sense to add `auth_db` as `dbname` property in the dataclass, but we would still need to use `POSTGRESQL_DEFAULT_DATABASE` from the constants.

src/utils/auth.py
theoctober19th: This is the most simplest implementation but we may add more logic specific to creation of admin user in the future (for example, storing the password in Juju secrets, etc.) for which a dedicated functional scope would be desirable. I agree at this point it is not needed, but I will keep it because it doesn't add any complexity either but rather makes the code more structured.

src/database.py
theoctober19th: I have updated the `DatabaseConnectionInfo` dataclass to include `dbname` as well. (This will be used by default if `dbname` param is not passed in the `execute` function. I believe that a table name is something that is passed in a query and thus would not have it as an attribute to DatabaseConnectionInfo (because this class should be able to query to / from multiple tables and not only one table.)

src/config/kyuubi.py
theoctober19th: Updated.

src/utils/auth.py
theoctober19th: I've commented about this [here](https://github.com/canonical/kyuubi-k8s-operator/pull/8#discussion_r1570005304)

src/utils/auth.py
theoctober19th: I removed the `AUTHENTICATION_DATABASE_NAME` as it will now be available as `database.dbname`, but these other constants are required. The `DEFAULT_ADMIN_USERNAME` is different from `database.username` because `DEFAULT_ADMIN_USERNAME` is the username of Kyuubi user while `database.username` is a PostgreSQL user.

tests/integration/test_charm.py
```diff
@@ -332,6 +337,215 @@ async def test_remove_postgresql_metastore_relation(ops_test: OpsTest, test_pod,
...
+    await ops_test.model.wait_for_idle(
+        apps=[APP_NAME, charm_versions.postgres.application_name], timeout=1000, status="active"
+    )
+
+    logger.info("Integrating kyuubi-k8s charm with postgresql-k8s charm over auth-db endpoint...")
+    await ops_test.model.integrate(charm_versions.postgres.application_name, f"{APP_NAME}:auth-db")
+
+    logger.info("Waiting for postgresql-k8s and kyuubi-k8s charms to be idle...")
+    await ops_test.model.wait_for_idle(
+        apps=[APP_NAME, charm_versions.s3.application_name], timeout=1000
```
welpaolo, line 353: **issue** you are looking for s3 app in idle instead of `postgresql`

welpaolo, line 353:
```suggestion
        apps=[APP_NAME, charm_versions.postgres.application_name], timeout=1000
```

tests/integration/test_charm.py
```diff
@@ -332,6 +337,215 @@ async def test_remove_postgresql_metastore_relation(ops_test: OpsTest, test_pod,
...
+
+    logger.info("Running action 'get-password' on kyuubi-k8s unit...")
+    action = await kyuubi_unit.run_action(
+        action_name="get-password",
+    )
+    result = await action.wait()
+    new_password = result.results.get("password")
+
+    assert new_password != old_password
+    assert new_password == password_to_set
```
welpaolo, line 505: After the change of password we need to check if the connection will succeed with the new password

tests/integration/test_charm.py
```diff
@@ -332,6 +337,215 @@ async def test_remove_postgresql_metastore_relation(ops_test: OpsTest, test_pod,
...
+            "table_444",
+        ],
+        capture_output=True,
+    )
+    print("========== test_jdbc_endpoint.sh STDOUT =================")
+    print(process.stdout.decode())
+    print("========== test_jdbc_endpoint.sh STDERR =================")
+    print(process.stderr.decode())
+    logger.info(f"JDBC endpoint test returned with status {process.returncode}")
+    assert process.returncode == 0
```
welpaolo, line 546: Here we need to check that login will failed with the old credentials

welpaolo (commented):
It looks a good starting point. The way the different component are structured need some improvement that probably we will address with the refactoring.

I left some comments to address

tests/integration/test_charm.py
theoctober19th: Good catch, thanks @welpaolo !

tests/integration/test_charm.py
theoctober19th: When the authentication in Kyuubi is removed, Kyuubi accepts all connections and doesn't check the credentials at all. That's why, the login would not fail with the old credentials as authentication is removed.

src/charm.py
```diff
@@ -170,6 +217,47 @@ def _on_get_jdbc_endpoint(self, event: ActionEvent):
...
+            event.fail("The action can be run only on leader unit")
+            return
+
+        auth = Authentication(self.auth_db_connection_info)
+        password = auth.generate_password()
+
+        if "password" in event.params:
+            password = event.params["password"]
+
+        if password == auth.get_password(DEFAULT_ADMIN_USERNAME):
```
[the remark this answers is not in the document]
deusebio: why not? It feels appropriate to me. I mean this is the main of the action. We can do these checks and return an error otherwise

tests/integration/test_charm.py
theoctober19th: Updated.

tests/integration/test_charm.py
theoctober19th: Updated.

src/charm.py
```diff
@@ -170,6 +214,47 @@ def _on_get_jdbc_endpoint(self, event: ActionEvent):
         result = {"endpoint": self.workload.get_jdbc_endpoint()}
         event.set_results(result)
 
+    def _on_get_password(self, event: ActionEvent) -> None:
+        """Returns the password for admin user."""
+        if not self.is_authentication_enabled():
+            event.fail(
+                "The action can only be run when authentication is enabled. "
+                "Please relate kyuubi-k8s:auth-db with postgresql-k8s"
```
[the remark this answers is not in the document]
theoctober19th: Updated.

src/charm.py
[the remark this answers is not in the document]
theoctober19th: I took reference from the `postgresql-k8s` in this part and included this check. I also think this is an appropriate place for this check, but would like to hear a different opinion.

src/utils/auth.py
theoctober19th: Updated.

src/utils/auth.py
deusebio: What about then to have this signature:

```
Authentication(database:DatabaseConnectionInfo, table_name: str)
```

This makes the class "pure", without using external, non-declared and implicit dependency. It should even make it easier to test

src/database.py
deusebio: What about then to have this signature:

```
Authentication(database:DatabaseConnectionInfo, table_name: str)
```

This makes the class "pure", without using external, non-declared and implicit dependency. It should even make it easier to test

src/utils/auth.py
deusebio: This was required because we were hardcoding the admin function (see other comment) or similar functions. We can make a call if this can/should be ever be customized, but if we want to do that I would make the "DEFAULT_ADMIN_USERNAME" a class static attribute of `Authentication`. Again, following the functional tech talk I was giving, the pattern of injecting constants like this here is not very "pure". Hence the suggestion

src/database.py
theoctober19th: I'm more comfortable with having these constants as static class properties rather than receiving them as constructor param. This way, the class would not have any external variable reference and this would not require us change other parts in the codebase as well since we keep the same constructor signature. This will kill two birds with a single stone. WDYT @deusebio ?

src/utils/auth.py
theoctober19th: I agree to make this a class static attribute. I'll update this.

src/charm.py
```diff
@@ -89,6 +110,30 @@ def _on_metastore_db_relation_removed(self, event) -> None:
...
+    def _on_auth_db_created(self, event: DatabaseCreatedEvent) -> None:
+        logger.info("Authentication database created...")
+        hostname, port = event.endpoints.split(":")
+        db_connection_info = DatabaseConnectionInfo(
+            endpoint=hostname,
+            username=event.username,
+            password=event.password,
+            dbname=AUTHENTICATION_DATABASE_NAME,
+        )
+        Authentication(db_connection_info).prepare_auth_db()
```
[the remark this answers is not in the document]
theoctober19th: I changed this to regular class. This class has only one attribute and does a lot of utility actions (like running queries, etc.) and hence I feel better that this be a regular class and not a dataclass.

src/config/spark.py
```diff
@@ -0,0 +1,87 @@
...
+from typing import Optional
+
+from lightkube import Client
+
+from constants import KYUUBI_OCI_IMAGE
+from s3 import S3ConnectionInfo
+from utils.logging import WithLogging
+
+
+class SparkConfig(WithLogging):
```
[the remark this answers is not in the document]
theoctober19th: Last time, we realized that Kyuubi would not read the properties tagged with a service account by default (since Kyuubi uses its own Spark binaries and not the binaries that wrap the Spark8t library).

Later we discussed about the possibility of Kyuubi reading the service account props directly from secrets, and then the charm preparing the file on-the-fly. 

I didn't include that as part of this ticket because I believe that constitutes the tasks we do during integration of Kyuubi with Integration Hub. I didn't want to extend the scope of this ticket at the moment and would like to pick this as part of Kyuubi <> Integration Hub integration task that we'll pick next week. Do you agree?


P.S. Even when we add logic to read the Spark props from Kubernetes secret, I think maybe the Kyuubi configuration and Hive configurations still need to be constructed by charm itself?

src/utils/auth.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python3
+
+# Copyright 2024 Canonical Limited
+# See LICENSE file for licensing details.
+
+"""S3 connection utility classes and methods."""
```
welpaolo, line 6: Please update description here

welpaolo (approved):
Looks good to me :) 

I left a couple of small comments.

On the open topic I will let @deusebio decide whenever we need to approach the update now or in a separate PR.

src/utils/auth.py
theoctober19th: Updated.

src/config/spark.py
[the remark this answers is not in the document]
deusebio: As you prefer, I'm also happy to address it in the next ticket, but I want this to land. I feel we are missing to centrilize code and business logic otherwise. 

> Later we discussed about the possibility of Kyuubi reading the service account props directly from secrets, and then the charm preparing the file on-the-fly.

Yes, that's what I meant. We won't put any spark-client wrapper around kyuubi, but we can use spark8t functionalities for creating the spark props file. 

> I think maybe the Kyuubi configuration and Hive configurations still need to be constructed by charm itself?

Yes, totally! These are Kyuubi specific right now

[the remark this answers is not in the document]
theoctober19th: I agree with you that this is the way we want to go forward with. It is just a matter of whether we do it during dedicated task for Kyuubi <> Integration Hub next week, or extend this and move this ticket to the next pulse. I prefer to address these during integration ticket we pick next, but I'm okay to address these here if you prefer it otherwise.

src/database.py
deusebio: I don't have strong opinion, but If I have to say my preference, I would make the table name configurable and admin to be a class static property. Such that one could even use this manager in tests easily with different tables or abstracted from the charm usage

src/config/spark.py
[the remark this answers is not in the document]
deusebio: I feel that it is a bit more general than just integration with Integration Hub. I mean, this is needed also when we integrate with a service account that was configured using spark-client. Hence the comment. 

I'm not sure is fully in scope for the ticket of integration hub, as the task is more an integration tests to be implemented in spark-k8s-bundle. When we integrate, we may have to fix UX elsewhere, but maybe this feature (meaning that kyuubi should be able to read configuration that a service account comes with) could already be implemented. 

Maybe the right call would be to add a ticket for this specifically if you like, that we can add to the pulse. It should not be a huge ticket, actually mostly replacing this class with the call above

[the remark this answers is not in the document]
deusebio: I have created [this](https://warthogs.atlassian.net/browse/DPE-4165) ticket. I believe it is fair to address this in a proper ticket. I have also provided a bit of description. Let's follow up on MM or on the ticket if there is unclarity. Resolving the comment now

.github/workflows/release.yaml
```diff
@@ -0,0 +1,31 @@
...
+    uses: canonical/data-platform-workflows/.github/workflows/build_charm_without_cache.yaml@v6.1.1
+
+  publish:
+    name: Release charm
+    needs:
+      - ci-tests
+      - build
+    uses: canonical/data-platform-workflows/.github/workflows/release_charm.yaml@v6.1.1
+    with:
+      channel: 1.9/edge
```
deusebio, line 26: can we start from `latest/edge` instead?

theoctober19th: Sure, updated.

deusebio (approved):
LGTM!

PR #9, opened 2024-04-22
[DPE-3937] Add kyuubi_client provider interface

theoctober19th:


deusebio (approved):
Looks very good! well done!

tests/integration/app-charm/src/charm.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python3
+# Copyright 2024 Ubuntu
```
welpaolo, line 2: Why do we have `Ubuntu` instead of `Canonical`?

welpaolo (approved):
LGTM

tests/integration/app-charm/src/charm.py
theoctober19th: Good catch, thanks. Just updated.

PR #14, opened 2024-05-08
[DPE-4252] Refactor Kyuubi into the new pattern

theoctober19th:


src/charm.py
```diff
@@ -55,8 +44,16 @@ class KyuubiCharm(ops.CharmBase):
...
+        # Workload
+        self.workload = KyuubiWorkload(
+            container=self.unit.get_container(KYUUBI_CONTAINER_NAME),
+        )
+
+        # Context
+        self.context = Context(self)
+
+        # Requirers
         self.metastore_db = DatabaseRequires(
```
theoctober19th, line 57: I have moved the `requires` objects to the main charm class, because in the current implementation, the `context` object holds all the state information (regarding metastore db information, auth db information ,s3 information and service account information). Reading the relation data from `Context` class is as straightforward for database relation as it is for s3 relation, because for database relation, the username and password are stored in the secrets. To view them in plaintext, `db_requires.fetch_relation_data()` is used. However, it is not straightforward for the `Context` class to access `db_requires` from inside the event handlers (I feel like it is anti pattern). That's why, I moved the requires relation definition to the charm class so that everyone interested can access it from there.

theoctober19th: And for consistency, I have moved S3 requires and Provider relation definition to the main class as well.

src/core/context.py
```diff
@@ -0,0 +1,73 @@
...
+                password=data["password"],
+                dbname=data["database"],
+                # dbname=METASTORE_DATABASE_NAME,
+            )
+        return None
+
+    @property
+    def auth_db(self):
+        """The state of authentication DB connection."""
+        for data in self.charm.auth_db.fetch_relation_data().values():
```
theoctober19th, line 54: ☝️ The context reads the relation data using `db_requires.fetch_relation_data()` here, just like we did before refactoring.

src/core/domain.py
```diff
@@ -0,0 +1,146 @@
...
+class DatabaseConnectionInfo:
+    """Class representing a information related to a database connection."""
+
+    endpoint: str
+    username: str
+    password: str
+    dbname: str
+
+
+class ServiceAccountInfo:
```
theoctober19th, line 127: At the moment, `ServiceAccountInfo` does not extend `StateBase` because these information are not read via relation but via config option until now. However, once we have full integration of Integration Hub to Kyuubi via relation, we can read these from the relation just like `S3ConnectionInfo`.

src/events/kyuubi.py
```diff
@@ -0,0 +1,68 @@
...
+
+        self.kyuubi = KyuubiManager(self.workload)
+
+        self.framework.observe(self.charm.on.install, self._on_install)
+        self.framework.observe(self.charm.on.kyuubi_pebble_ready, self._on_kyuubi_pebble_ready)
+        self.framework.observe(self.charm.on.update_status, self._update_event)
+        self.framework.observe(self.charm.on.config_changed, self._on_config_changed)
+
+    @compute_status
+    def _on_install(self, event: ops.InstallEvent) -> None:
```
theoctober19th, line 35: The decorator `compute_status` will take care of setting appropriate status at the end of hook, that's why the function definition is not needed.

src/charm.py
deusebio: This is actually my main concern. This does not quite follow the pattern. The idea here is that we should not use the `DatabaseRequires` anymore, but swap to use the `*Data` classes and `*EventsHandlers` classes. 

The `Data` classes should in general be wrapped by the context, whereas the events handlers should be imported and used in the events classes. 

See [here](https://github.com/canonical/spark-integration-hub-k8s-operator/blob/main/src/core/context.py#L25) and [here](https://github.com/canonical/karapace-operator/blob/main/src/core/cluster.py#L42) for the context. And [here](https://github.com/canonical/karapace-operator/blob/main/src/events/kafka.py#L32) for the events handlers.

src/workload/k8s.py
deusebio, line 1: I would move these into common folder. 

The reason is that I would follow this pattern:

* common => submodule with product independent libraries
* core => submodle with data models and domain objects (no business logic - only IO parsing of data bags)
* managers => submodule with product specific business logic (possibly based on core modules)
* events => submodule with product specific charm logic 

@welpaolo what do you think?

src/workload/kyuubi.py
deusebio, line 1: I'm actually not sure about where to place this one. I have historically put this in the `core` module, but it is tricky as this may contain business logic and not domain objects. However it is clear that managers should depend on this. And general the dependency rule can be 

common (product independent) < core (product core modules) < managers (business logic) < events (charm logic)

The point is that this class is clearly product specific. So the question is whether it should be placed in core or in managers. I don't think it would be worth creating a new "class" of submodules between core and managers. Of all of these options, I would maybe stick to placing it in core. But feel free to chip in with your thoughts.

src/events/metastore.py
```diff
@@ -0,0 +1,59 @@
...
+from managers.kyuubi import KyuubiManager
+from utils.logging import WithLogging
+from workload.base import KyuubiWorkloadBase
+
+
+class MetastoreEvents(BaseEventHandler, WithLogging):
+    """Class implementing PostgreSQL metastore event hooks."""
+
+    def __init__(self, charm: CharmBase, context: Context, workload: KyuubiWorkloadBase):
+        super().__init__(charm, "s3")
```
welpaolo, line 23:
```suggestion
        super().__init__(charm, "metastore")
```

src/events/metastore.py
```diff
@@ -0,0 +1,59 @@
...
+            auth_db_info=self.context.auth_db,
+            service_account_info=self.context.service_account,
+        )
+
+    @compute_status
+    def _on_metastore_db_relation_removed(self, event) -> None:
+        self.logger.info("Mestastore database relation removed")
+        self.kyuubi.update(
+            s3_info=self.context.s3,
+            metastore_db_info=self.context.metastore_db,
```
welpaolo, line 56:
```suggestion
            metastore_db_info=None,
```

welpaolo: I believe that here, you will need to update `metastore_db` to `None`

src/events/s3.py
```diff
@@ -0,0 +1,57 @@
...
+            auth_db_info=self.context.auth_db,
+            service_account_info=self.context.service_account,
+        )
+
+    @compute_status
+    def _on_s3_credential_gone(self, _: CredentialsGoneEvent):
+        """Handle the `CredentialsGoneEvent` event for S3 integrator."""
+        self.logger.info("S3 Credentials gone")
+        self.kyuubi.update(
+            s3_info=self.context.s3,
```
welpaolo, line 53:
```suggestion
            s3_info=None,
```

src/workload/k8s.py
welpaolo: I agree

welpaolo (commented):
Good starting point but please take a look at the comments because some of them maybe problematic

src/charm.py
theoctober19th: I agree with this one. I'll refactor this so that we use `*Data` and `*EventHandler` classes separately.

src/charm.py
```diff
@@ -66,273 +63,17 @@ def __init__(self, *args):
...
-                password=data["password"],
-                dbname=AUTHENTICATION_DATABASE_NAME,
-            )
-        return None
-
-    def _on_s3_credential_changed(self, _: CredentialsChangedEvent):
-        """Handle the `CredentialsChangedEvent` event from S3 integrator."""
-        logger.info("S3 credentials changed")
-        self.update_service()
+        self.s3 = S3Requirer(self, S3_INTEGRATOR_REL)
```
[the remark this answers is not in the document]
theoctober19th: This was moved here for consistency earlier because I had to move `DatabaseRequires` to the top level earlier. However, since I am going to split `DatabaseRequires` to `*Data` and `*EventHandler` implementations separately, I'll move this down the line to event handlers.

src/events/metastore.py
theoctober19th: `self.context.metastore_db` would return `None` by itself when the relation is not there, so it is effectively the same thing. Also, this would make the updates consistent (we can maybe move this common part up the inheritance chain in the future to reduce code redundancy).

src/events/s3.py
theoctober19th: `self.context.s3` would return `None` by itself when S3 relation has broken.

src/core/context.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env python3
+# Copyright 2024 Canonical Limited
+# See LICENSE file for licensing details.
+
+"""Charm Context definition and parsing logic."""
+
+
+# from charms.data_platform_libs.v0.data_interfaces import RequirerData
```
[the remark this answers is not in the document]
theoctober19th: Removed.

src/events/actions.py
```diff
@@ -0,0 +1,77 @@
...
+
+        self.kyuubi = KyuubiManager(self.workload)
+        self.auth = AuthenticationManager(self.context.auth_db)
+
+        self.framework.observe(self.charm.on.get_jdbc_endpoint_action, self._on_get_jdbc_endpoint)
+        self.framework.observe(self.charm.on.get_password_action, self._on_get_password)
+        self.framework.observe(self.charm.on.set_password_action, self._on_set_password)
+
+    def _on_get_jdbc_endpoint(self, event: ActionEvent):
+        result = {"endpoint": self.workload.get_jdbc_endpoint()}
```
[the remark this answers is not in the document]
theoctober19th: Updated.

src/workload/k8s.py
theoctober19th: Moved the file to `common` folder.

src/workload/kyuubi.py
theoctober19th: Moved the file to `core` folder.

src/core/context.py
```diff
@@ -0,0 +1,85 @@
...
+    S3_INTEGRATOR_REL,
+)
+from core.domain import DatabaseConnectionInfo, S3ConnectionInfo, ServiceAccountInfo
+from utils.logging import WithLogging
+
+
+class Context(WithLogging):
+    """Properties and relations of the charm."""
+
+    def __init__(self, charm: CharmBase):
```
deusebio, line 24: I've been checking whether we could use some less wider object, other than charm in the constructor. The reason is that charm is a very wide-scoped object. I would prefer to provide more scoped inputs, like `config` and `model`. 

I believe we can do that as long as the `ServiceAccountInfo` only depends on config. All the other usage in the class would only need `model`, and not `charm`.

src/events/kyuubi.py
```diff
@@ -0,0 +1,71 @@
...
+        self.kyuubi.update(
+            s3_info=self.context.s3,
+            metastore_db_info=self.context.metastore_db,
+            auth_db_info=self.context.auth_db,
+            service_account_info=self.context.service_account,
+        )
+
+    @compute_status
+    def _update_event(self, event):
+        """Handle the update event hook."""
```
deusebio, line 60: **niptick** I would just add the `pass` here.

deusebio (approved):
Great work! The code looks very nice and really closely follow the new pattern! Well done!

src/core/context.py
theoctober19th: Updated the `Context` constructor to receive `model` and `config` instead of `charm`.

src/events/kyuubi.py
theoctober19th: Updated.

welpaolo (approved):
LGTM. Great work!
I left some minor comments for the addition of docstring

PR #19, opened 2024-05-30
[DPE-4472] Pin the same version of rock image everywhere in Kyuubi charm

theoctober19th:


deusebio:
not a bit fan of this. I mean: when we update this we need to remember to update 3 places. Can we put some logic to fetch this info directly from metadata.yaml?

welpaolo:
I agree can you try to find an automatic way to propagate this info ?

.github/workflows/ci.yaml
```diff
@@ -111,7 +111,7 @@ jobs:
         name: Cache Spark Image Locally
         run: |
           # Download image for avoiding time out
-          IMAGE="ghcr.io/canonical/charmed-spark:3.4-22.04_edge"
+          IMAGE="ghcr.io/canonical/charmed-spark:3.4-22.04_edge@sha256:90a1f26a33d7524b00b5b0692f03ed96919ea89ded571943d2b5aa7802389f87"
```
deusebio, line 114: This one is easy:

```
yq ".resources.kyuubi-image.upstream-source" metadata.yaml
```

tests/integration/setup/testpod_spec.yaml
```diff
@@ -7,5 +7,5 @@ metadata:
   name: testpod
 spec:
   containers:
-  - image: ghcr.io/canonical/charmed-spark:3.4-22.04_edge
+  - image: ghcr.io/canonical/charmed-spark:3.4-22.04_edge@sha256:90a1f26a33d7524b00b5b0692f03ed96919ea89ded571943d2b5aa7802389f87
```
deusebio, line 10: We can replace values on the way with [this](https://github.com/canonical/charmed-spark-rock/blob/3.4-22.04/edge/tests/integration/integration-tests.sh#L136), with no need to pin in the yaml directly

theoctober19th:
Yes, I can definitely do that. Would there possibly be cases where we would want the pod image to be different from charm image (eg, possibly some changes in spark k8s toolkit we may require in Pod, but we are not ready to pin the new revision in charm metadata just yet)?

deusebio:
uhm, not sure. But if any, we will revisit this. But even then, pinning would be quite good anyway such that we don't break CI when releasing new images.

tests/integration/setup/testpod_spec.yaml.template
```diff
@@ -7,5 +7,5 @@ metadata:
   name: testpod
 spec:
   containers:
-  - image: ghcr.io/canonical/charmed-spark:3.4-22.04_edge
+  - image: ${kyuubi_image}
```
deusebio, line 10: **praise** This is very nice!

deusebio (approved):
Looks great!

welpaolo (approved):

PR #22, opened 2024-06-26
[DPE-4485] Use charmed-spark-kyuubi OCI image

deusebio:


src/constants.py
```diff
@@ -20,7 +20,7 @@
 
 JDBC_PORT = 10009
 
-KYUUBI_OCI_IMAGE = "ghcr.io/canonical/charmed-spark:3.4-22.04_edge"
+KYUUBI_OCI_IMAGE = "ghcr.io/canonical/charmed-spark-kyuubi:3.4-22.04_edge"
```
theoctober19th, line 23: suggestion: can we get this value from `metadata.yaml` somehow so that we have only one place where we specify the image (i.e. metadata.yaml) and everything else works out of the box?

theoctober19th (approved):
LGTM. I've left a nice-to-have suggestion, however it is not a blocking change.

PR #26, opened 2024-08-15
Draft: Zookeeper Integration

theoctober19th:


metadata.yaml
```diff
@@ -44,6 +44,9 @@ requires:
   spark-service-account:
     interface: spark_service_account
     limit: 1
+  zookeeper:
+    interface: zookeeper
```
marcoppenheimer, line 48: **nitpick:** I think this is `zookeeper_client`

theoctober19th: I took reference from the Kafka code base: https://github.com/canonical/kafka-k8s-operator/blob/main/metadata.yaml#L57.

src/config/kyuubi.py
```diff
@@ -41,9 +51,22 @@ def _auth_conf(self) -> dict[str, str]:
...
 
+    @property
+    def _ha_conf(self) -> dict[str, str]:
+        if not self.zookeeper_info:
+            return {}
+        return {
+            "kyuubi.ha.addresses": self.zookeeper_info.uris,
+            # FIXME: Get this value from self.context.zookeeper.uris when znode created by
+            # zookeeper charm has enough permissions for Kyuubi to work
+            "kyuubi.ha.namespace": HA_ZNODE_NAME,
```
theoctober19th, line 62: Here, as a workaround we supply a different name to DatabaseRequires and a different name to Kyuubi config so that Kyuubi can create the znode itself with the necessary permissions.

Once we figure out the ALCs that are set to znode creates by the Zookeeper charm and agree on it, we can then use the znode created by zookeeper instead.

src/core/context.py
```diff
@@ -36,6 +43,13 @@ def __init__(self, model: Model, config: ConfigData):
             extra_user_roles="superuser",
         )
 
+        # FIXME: The database_name currently requested is a dummy name
+        # This should be replaced with the name of actual znode when znode created
+        # by zookeeper charm has enough permissions for Kyuubi to work
+        self.zookeeper_requirer_data = DatabaseRequirerData(
+            self.model, ZOOKEEPER_REL, database_name=HA_ZNODE_NAME_TEMP
```
theoctober19th, line 50: As a workaround, we ask Zookeeper to just create a dummy node that we won't use (because this znode does not have enough ACL and permissions for HA to work).

metadata.yaml
deusebio: Nope, I believe we kept `zookeeper` since that was the initial name in Kafka, and changing the interface name would have been an huge pain (not backward compatible). Because of this, even the name in the [interface repository](https://github.com/canonical/charm-relation-interfaces/pull/144)

src/config/kyuubi.py
deusebio: yes, we should probably update the relation interface to allow to create "databases" that are writable by the requirer, and readable by others. This seems a fairly common use-case we may want to have a consistent UX, and beside, since we are using the `DatabaseRequires/DatabaseProvides` classes, we should definitely discuss with the wider DP team. 

I have mentioned this to @welpaolo today that I feel it is worth bringing this up at DP Office hours

src/constants.py
```diff
@@ -5,22 +5,30 @@
...
 JDBC_PORT = 10009
-
 KYUUBI_OCI_IMAGE = "ghcr.io/canonical/charmed-spark-kyuubi:3.4-22.04_edge"
-
 DEFAULT_ADMIN_USERNAME = "admin"
-KYUUBI_CLIENT_RELATION_NAME = "jdbc"
+
+# Zookeeper literals
+HA_ZNODE_NAME = "/kyuubi"
+HA_ZNODE_NAME_TEMP = "/kyuubi-temp"
```
deusebio, line 34: **suggestion** I would not put tmp constants (which are also workaround) among the actual constants. Given it is a workaround, we could just hard code this in the request (or even randomly generate that) :D 

anyhow, it is a fairly small point

src/events/actions.py
```diff
@@ -39,15 +39,22 @@ def _on_get_jdbc_endpoint(self, event: ActionEvent):
...
-        result = {"endpoint": self.workload.get_jdbc_endpoint()}
+
+        if self.context.is_ha_enabled():
+            address = self.context.zookeeper.uris
+            # FIXME: Get this value from self.context.zookeeper.uris when znode created by
+            # zookeeper charm has enough permissions for Kyuubi to work
+            namespace = HA_ZNODE_NAME
+            if not address.endswith("/"):
+                address += "/"
+            endpoint = f"jdbc:hive2://{address};serviceDiscoveryMode=zooKeeper;zooKeeperNamespace={namespace}"
```
deusebio, line 53: **suggestion** I would put this business logic more in the `kyuubi` manager, rather than in the `handlers`. 

We could probably just expose a method in the manager that is `get_endpoint`, and there place the business logic, also based on the HA/non-HA logic

deusebio (approved):
The code looks good to me! I just have a couple of comments/suggestions, but they are not blocking

src/constants.py
theoctober19th: I agree.

theoctober19th:
@deusebio Thanks a lot for the review! I agree with your points. Since we're going to merge this to the feature/kyuubi-ha branch, I'd rather piggyback these non-blocking suggestions in the next PR I open in the same feature (that way, the CI need not be triggered again). At the end, these will appear in `feature/kyuubi-ha` anyway.

I'm waiting for @welpaolo's review here and once I have that, I'll merge this and push your suggestions in the next PR in the same feature/epic.

welpaolo (approved):
LGTM, I left a couple of minor comments

tests/integration/conftest.py
```diff
@@ -160,7 +171,17 @@ def test_pod():
...
+        [
+            "kubectl",
+            "wait",
+            "--for",
+            "condition=Ready",
+            f"pod/{pod_name}",
+            "-n",
+            namespace,
+            "--timeout",
+            "60s",
```
welpaolo, line 183: Can you increase timeout to avoid failed test in case of network issue? I will suggest to put 300 sec

welpaolo (approved):

tests/integration/conftest.py
theoctober19th: The CI pre-downloads the image and caches it before even running the integration tests. Also, since this code exists from long ago and isn't something added now, I'll leave it as such for now. During improving flaky tests for Spark K8s bundle, I can come back to it if it is really causing issues.

theoctober19th:
I'm merging this now to `feature/kyuubi-ha`. The suggestions will be shipped to this feature branch in next PR.

PR #28, opened 2024-08-27
[Draft]: HA Tests for KYuubi

theoctober19th:


deusebio (commented):
The PR looks code and the tests seem sane to me, and I don't have many concerns on the actual business logic of the tests.

I just feel that here we could have benefited from starting to include the `spark_test` library in Kyuubi. In the PR, I saw a bunch of "general" code (using kyuubi, deleting pods, gathering information, creating buckets, etc) that could have been included in the spark_test or we could have used abstraction from there. 

I personally don't believe we will ever refactor entirely the test suite to use `spark_test` in one go, but rather we could probably do that progressively and iteratively (but let me know if you feel otherwise). But in this sense, when creating a good bunch of tests (as in this case), that could be a good occasion to start importing the library (and updating the library) instead of keeping increasing the code base in the charm (which it will just increase the hassle and complexity of importing `spark_test`)

Let's discuss this and how we want to progressively importing the `spark_test` if we are still keen on doing so

tests/integration/helpers.py
```diff
@@ -0,0 +1,289 @@
...
+    )
+    print("========== test_jdbc_endpoint.sh STDOUT =================")
+    print(process.stdout.decode())
+    print("========== test_jdbc_endpoint.sh STDERR =================")
+    print(process.stderr.decode())
+    logger.info(f"JDBC endpoint test returned with status {process.returncode}")
+    return process.returncode == 0
+
+
+# async def get_kyuubi_znode_children(ops_test: OpsTest, test_pod) -> list[str]:
```
welpaolo, line 73: Do we need to keep this commented function?

welpaolo (approved):
LGTM! Great start!
Please add docstring for helper functions.

src/core/context.py
```diff
@@ -47,7 +46,7 @@ def __init__(self, model: Model, config: ConfigData):
         # This should be replaced with the name of actual znode when znode created
         # by zookeeper charm has enough permissions for Kyuubi to work
         self.zookeeper_requirer_data = DatabaseRequirerData(
-            self.model, ZOOKEEPER_REL, database_name=HA_ZNODE_NAME_TEMP
+            self.model, ZOOKEEPER_REL, database_name="/kyuubi-test"
```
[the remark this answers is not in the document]
theoctober19th: This comment exists just 1 line earlier.

tests/integration/helpers.py
theoctober19th: We don't. However, I kept this here because I wanted both of your and @deusebio's view on whether we should use the uncommented (`kyuubi-ctl` way) of fetching this information or this commented `zookeeper` way. To me, `ls znode` approach using zookeeper is way cleaner and straightforward, but I equally understand your preference over using `kyuubi` native tools to achieve the same result. @deusebio, what do you think?

deusebio (approved):

tests/integration/helpers.py
deusebio: Uhm, I don't have strong opinions. Using the `kyuubi-ctl` is interesting since we would ALSO test that. But if you prefer to just use `kazoo` since this function is kyuubi independent, this makes sense too. 

Up to you really. But I would not use commented code, why not using a flag in the function `backend` and we provide BOTH implementation out of the box?

PR #29, opened 2024-09-02
[DPE-4481] Handle cases where charm is deployed without using the --trust flag

theoctober19th:


.github/workflows/ci.yaml
```diff
@@ -84,7 +85,7 @@ jobs:
           provider: microk8s
           channel: 1.28-strict/stable
           microk8s-group: snap_microk8s
-          microk8s-addons: "hostpath-storage dns minio metallb:${{ env.IPADDR }}-${{ env.IPADDR }}"
+          microk8s-addons: "rbac hostpath-storage dns minio metallb:${{ env.IPADDR }}-${{ env.IPADDR }}"
```
theoctober19th, line 88: Here, I've enabled `rbac` in our test K8s environment because without RBAC, there's no point testing cluster permissions.

deusebio: Indeed! Good catch! I believe we  should always test with RBAC enabled! Thanks!

deusebio (approved):
LGTM! Thanks!

welpaolo (approved):
LGTM

PR #31, opened 2024-09-09
[DPE-5001] Allow Kyuubi to work without specifying namespace 

welpaolo:
This PR improves the Kyuubi charm by letting it running without specifying the `namespace` configuration. This will let Kyuubi to run in the juju model namespace


pyproject.toml
```diff
@@ -37,7 +37,9 @@ pyyaml = "^6.0.1"
 boto3 = "^1.26.75"
 jsonschema = "^4.19.1"
 lightkube = "^0.15.2"
-psycopg2 = "^2.9.9"
+psycopg2-binary = "^2.9.9"
```
theoctober19th, line 40: question: Why do we replace `psycopg2` with `psycopg2-binary` here? AFAIK, Postgres recommends `psycopg2` be built from source in place of `psycopg2-binary`.

theoctober19th (commented):
Thanks a lot! This will save a lot of time for us from situations where we had to delete apps altogether because integration hub went into crashing state.

I have a small question regarding the change from `psycopg2` to `psycopg2-binary`.

deusebio (approved):
LGTM

PR #32, opened 2024-09-10
[DPE-5416] Update test S3 credentials generated by MinIO

theoctober19th:


welpaolo (approved):
LGTM
Minor: in my opinion we can keep the old logic comment for reference in case the K8s team will add back the old logic. @deusebio What's your opinion ?

deusebio (approved):
@welpaolo I don't mind too much, as you guys prefer. Also finding this PR is not that difficult either, but in general it depends whether this is going to be solved or not, and whether this PR is a workaround or is going to stay

PR #36, opened 2024-09-17
[DPE-5051] Loki logging

juditnovak:
This change is a result of a long research. Below are the findings.

# Failed attempts

## 1. Kyuubi server logs -- collected by Pebble.

    Ideally for such purpose we should use `charms.loki_k8s.v1.loki_push_api.LogForwarder`. 
    This is supposed to "inject a Pebble layer in each workload container the charm has access to, to configure Pebble's log forwarding feature and start sending logs to Loki."
    Goes as the following:
```
self._log_forwarder = LogForwarder(self, relation_name=COS_LOG_RELATION_NAME)
```
    However, the only workload that is being inserted is Loki's own workload...
    
    image
    image

    No other option but the Loki charm :-(
    
    image

    Once more reliable, this should be our reference point.

## 2. Kyuubi worker logs -- v1 `LogProxyConsumer`
     A Loki requirer relation can only take ONE particular source. Pebble or local logs -- but not _both_.
     Thus I was wondering to have two logger connections:
      - one for server logs (Pebble)
      - a 2nd one for worker logs (log files)
     While this is supposed to be possible technically, I was not able to publish a single log file using `v1`. 
     (Tried from multiple locations, multiple log files, ports, no success.)
     
# Success

## 3. `v0`, worker logs
    So the one thing that worked were simple server logs using the deprecated `v0` method. See PR.


tests/integration/test_charm.py
```diff
@@ -620,424 +696,365 @@ async def test_set_password_action(ops_test: OpsTest, test_pod):
...
+#
+#     expected_executor_pod_names = [
+#         driver_pod_name.replace("driver", "exec-1"),
+#         driver_pod_name.replace("driver", "exec-2"),
+#         driver_pod_name.replace("driver", "exec-3"),
+#     ]
+#
+#     assert len(executor_pod_names) == len(expected_executor_pod_names)
+#     assert set(executor_pod_names) == set(expected_executor_pod_names)
+
```
juditnovak, line 1052: [IMPORTANT] Don't forget to uncomment client tests once they are fixed on the pipeline.

deusebio (commented):
Looks good! 

I only have one relevant point:

Would it be possible to overload the same relation with both for the server and the spark entrypoint (see ps below)? It feels strange to have two different endpoints, and in particular one only for the entrypoint (note that here we are not talking about driver/executor, I would expect the entrypoints logs to be rather tiny). Have you tried to still have v0 and v1, and let them co-exists on the same endpoint? What is the COS team guidance on this? If this is not possible, but it will be possible in the future, maybe I would just forget about the worker entrypoint logs for now, and add it as the ecosystem allows

@welpaolo @theoctober19th what do you guys think?


ps. I have realized that these are actually not the worker logs, but probably only the spark entrypoint process.

theoctober19th (approved):
The logs in the /opt/kyuubi/work/... are indeed the spark-submit logs. However, I don't have strong reasons to not collect these logs. The PR looks good to me. The tests are to be uncommented, though.

juditnovak:
@deusebio 
It's a very good point. I thought I tried this and run into issues.
But now I've got a healthy pipeline proving the concenpt: https://github.com/canonical/kyuubi-k8s-operator/actions/runs/11037229789 (using commit  [Reloading Relation](https://github.com/canonical/kyuubi-k8s-operator/pull/36/commits/c4e4fdcd238ba252005efd54fbd55452bf78ea6e))

metadata.yaml
```diff
@@ -51,6 +51,12 @@ requires:
   zookeeper:
     interface: zookeeper
     limit: 1
+  logging:
+    interface: loki_push_api
+    limit: 1
+  # logging-worker:
+  #   interface: loki_push_api
+  #   limit: 1
```
deusebio, line 59: Let's remove this then! :D

deusebio:
Lovely! Very glad it worked out!!!! Thanks @juditnovak !

juditnovak:
In fact, given the amount of issues we're having with the worker logs, I'm thinking to split the PR in two (accoridng to the original tickets).

This one would remain with server logs + corresp tests (easy merge), and we can further investigate worker logs on a separate PR.

deusebio:
That's fine by me! I would just not call them worker logs, maybe more spark-process logs :D

metadata.yaml
juditnovak: Done as part of a full cleanup.

juditnovak:
PR cleaned up, Spark logs moved to https://github.com/canonical/kyuubi-k8s-operator/pull/39 where we could further discuss.

src/constants.py
```diff
@@ -23,6 +23,8 @@
 
 COS_METRICS_PORT = 10019
 COS_METRICS_PATH = "/metrics"
+COS_LOG_RELATION_NAME_SERVER = "logging"
+COS_LOG_RELATION_NAME_WORKER = "logging-worker"
```
deusebio, line 27: **niptick** we should probably remove this, right?

deusebio (approved):
Just a niptick! Thanks!

tests/integration/test_charm.py
juditnovak: All tests are enabled again. Related issues to be discussed on the dedicated PR https://github.com/canonical/kyuubi-k8s-operator/pull/39

juditnovak:
Hmmmm....

The same issue without worker logs publishing... :-/ (2 attempts consistentyl...)
https://github.com/canonical/kyuubi-k8s-operator/actions/runs/11040282729/job/30670447723

Switching the order of the tests, but "faking" a green pipeline doesn't change the fact that something is wrong here :-(

juditnovak:
Huh. :green_circle: 

The winnter stragegy, faking that "all works":
 - COS relation is set at the end of the test module
 - 10 times 60s `tenacity` retry, until we can finally pick up some server logs

theoctober19th:
IMO, the current version of test where we check some logs are ingested into Loki is fine for validating the logging with Loki is working. I'm happy to get this merged without checking for the "Starting Kyuubi server" log specifically, and even if we want to assert a specific log in the loki logs, we can assert some other logs that are ingested into Loki.

Moreover, the "Starting Kyuubi Server" log can be regenered if some config is changed in Kyuubi (becausre that would trigger the pebble Kyuubi layer to be restarted and thus the Server is started again. (that is, if this log is super important to be asserted, which I think we can proceed without).

theoctober19th (approved):

welpaolo (approved):

PR #38, opened 2024-09-23
[DPE-4349] Feature: Kyuubi High Availability with Zookeeper

theoctober19th:
This PR wants to merge everything we have done for the Kyuubi HA feature from feature branch to `main` branch.


deusebio (approved):
Looks very good! I would just rebase the branch, sqaush the last commits about fixing tests with the ticket number and this PR description, and then do the "rebase and merge" to keep the reference to the incremental work

src/constants.py
```diff
@@ -5,25 +5,32 @@
...
 COS_METRICS_PORT = 10019
 COS_METRICS_PATH = "/metrics"
-
 NAMESPACE_CONFIG_NAME = "namespace"
 SERVICE_ACCOUNT_CONFIG_NAME = "service-account"
 
+# Literals related to Kyuubi
 JDBC_PORT = 10009
-
 KYUUBI_OCI_IMAGE = "ghcr.io/canonical/charmed-spark-kyuubi:3.4-22.04_edge"
```
welpaolo, line 32: Do we want to pin a specific version of the image ?

welpaolo (approved):
Looks good to me! 
I left a minor comment that could also be addressed in the future

src/constants.py
theoctober19th: Done. :heavy_check_mark: :)

PR #44, opened 2024-10-04
[DPE-5528]  Kyuubi upgrade

welpaolo:
This PR introduces the feature of upgrades to the Kyuubi charm.

This ticket includes both the functionality and the relative integration tests. 

**_NOTE:_**  Eventually, It ended up it was not needed to deregister the upgraded Kyuubi unit from the znode in the Zookeeper cluster since the heartbeat mechanism implemented in Kyuubi ended up to be really efficient in the detection on missing units and to perform the corresponding update in the znode.


theoctober19th (commented):
Good work, Paolo! Looks good to me.

theoctober19th (approved):

deusebio (commented):
That's pretty neat! 

I only have one question: How are we going around about stopping the Kyuubi service should there be connections to it?

welpaolo:
> That's pretty neat!
> 
> I only have one question: How are we going around about stopping the Kyuubi service should there be connections to it?

In the current implementation the connection between a Kyuubi unit and  connected clients is closed and the update is performed. The assumption is that the client should have re-connection policies to reconnect to the cluster.

deusebio (approved):

PR #46, opened 2024-11-19
[DPE-6000] Expose Kyuubi externally via K8s service

theoctober19th:
This PR adds the feature to expose Kyuubi externally as a K8s service.


theoctober19th:
@deusebio @welpaolo I'd like to have the approach here reviewed by you. Once we agree on the overall approach, I'll add the integration tests and open the PR for a detailed review.

src/events/base.py
```diff
@@ -68,6 +70,13 @@ def get_app_status(  # noqa: C901
         if self.charm.app.planned_units() > 1 and not self.context.zookeeper:
             return Status.MISSING_ZOOKEEPER.value
 
+        if self.charm.config.get("expose-external", "false") not in VALID_EXPOSE_EXTERNAL_VALUES:
+            return Status.INVALID_EXPOSE_EXTERNAL.value
+
+        service_util = ServiceUtil(self.charm.model)
```
theoctober19th, line 76: When computing the status (which happens immediately after `config-changed` and all other major hooks), we check the connectivity of the service endpoints and put the charm in Maintenance state. Later on when the deferred config-changed hook or any other hook uses `@compute_status` decorator, the connectivity is checked again and if found connectible, charm is put into active state again.

src/events/kyuubi.py
```diff
@@ -61,6 +71,18 @@ def _on_config_changed(self, event: ops.ConfigChangedEvent) -> None:
             zookeeper_info=self.context.zookeeper,
         )
 
+        # Check the newly created service is connectable
+        if not self.service_util.is_service_connectable():
```
theoctober19th, line 75: On `config-changed` event, we reconcile the required external service and immediately check whether the service is available for connection. If not, we simply defer this config-changed event in the hopes that when the deferred event comes back later, the service endpoint will be available.

It is to be noted that when the deferred event comes, `recoincile_services` will not delete and recreate the external service, it just checks and finds that the desired service is already created and return.

src/managers/kyuubi.py
```diff
@@ -32,14 +32,49 @@ def update(
...
         """Update Kyuubi service and restart it."""
-        spark_config = SparkConfig(
+        restart_required = False
+
+        new_spark_config = SparkConfig(
             s3_info=s3_info, service_account_info=service_account_info
         ).contents
-        hive_config = HiveConfig(db_info=metastore_db_info).contents
-        kyuubi_config = KyuubiConfig(db_info=auth_db_info, zookeeper_info=zookeeper_info).contents
+        try:
```
theoctober19th, line 40: Improving the way we update the configuration files inside the workload, we now attempt to update the configuration files and restart the container only if something has changed in the configuration from what is already inside the container. This will greatly improve the performance when `config-changed` event is triggered multiple times.

deusebio (commented):
I don't have any major concern with the business logic about creating services and deferring the event. There are a few things to be improved:

1. Validation logic for configuration options to comply with the standard and also avoid mixing business logic
2. general code improvements (see the comments attached)

But I believe it would be good to also have a cross-review, with @Batalex review here and @theoctober19th please have a look at the [PR being raised in ZooKeeper](https://github.com/canonical/zookeeper-k8s-operator/pull/116/files)

config.yaml
```diff
@@ -9,4 +9,12 @@ options:
...
     type: string
     description: The service account to be used by driver to create executor pods.
-    default: 'kyuubi-spark-engine'
\ No newline at end of file
+    default: 'kyuubi-spark-engine'
+  expose-external:
+    type: string
+    description: The mode in which the service should be exposed externally. Valid values are false, nodeport and loadbalancer.
+    default: 'false' 
+  loadbalancer-extra-annotations:
```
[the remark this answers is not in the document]
theoctober19th: This is not currently used, but when this PR is ready for full review, I will add the labels here to the LoadBalancer creation logic.

src/utils/service.py
```diff
@@ -0,0 +1,203 @@
...
+        )
+
+    def create_service(self, service_type, owner_references):
+        """Create the Kubernetes service with desired service type."""
+        desired_service = lightkube.resources.core_v1.Service(
+            metadata=lightkube.models.meta_v1.ObjectMeta(
+                name=self.service_name,
+                namespace=self.namespace,
+                ownerReferences=owner_references,  # the stateful set
+                labels={"app.kubernetes.io/name": self.app_name},
```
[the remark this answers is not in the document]
theoctober19th: This is not currently used, but when this PR is ready for full review, I will add the labels here to the LoadBalancer creation logic.

src/common/workload/k8s.py
```diff
@@ -46,7 +46,7 @@ def read(self, path: str) -> list[str]:
             raise FileNotFoundError
 
         with self.container.pull(path) as f:
-            return f.read().split("\n")
+            return f.read()
```
Batalex, line 49: **thought**: not saying that this is what's good for this code base, but kafka & zookeeper just return an empty list, which is good enough for our uses.  I am also curious as to why you felt the need the return the str content instead of keeping the splitlines.

We could get rid of a few try excepts later by returning the empty list

welpaolo (commented):
The approach looks good to me! 
Some part of the code can be simplified, and some nice addition like structured config can be added. 
Please continue in this direction and add integration tests

src/utils/service.py
```diff
@@ -0,0 +1,203 @@
...
+    @property
+    def _host(self) -> str:
+        """K8s service hostname for Kyuubi."""
+        # Example: kyuubi-k8s-service.my-model.svc.cluster.local
+        return f"{self.service_name}.{self.model_service_domain}"
+
+    def _get_node_host(self) -> set[str]:
+        """Return the node ports of nodes where units of this app are scheduled."""
+
+        def _get_node_address(node) -> str:
```
Batalex, line 53: **suggestion**: we could inline this function content, since we do not pass it to a `map()` or something like that

tests/unit/test_charm.py
```diff
@@ -34,7 +34,8 @@ def parse_kyuubi_configurations(tmp_path: Path) -> dict[str, str]:
         )
 
 
-def test_start_kyuubi(kyuubi_context):
+@patch("lightkube.Client")
```
Batalex, line 37: **suggestion**: if we need to patch this everywhere, maybe we could make it a fixture with `autouse=True` in `conftest.py`

Batalex (commented):
Left a few comments. Overall, the implementation is quite similar to ZK's with a notable exception:  
here we "look before we leap" by looking for the service before doing anything, whereas ZK applies the k8s without any question. I don't have a strong enough knowledge of K8s (and it's various implementations) to know if blindly applying the service will be an issue. So far, it's not.

src/common/workload/k8s.py
theoctober19th: Previously the `read` would return list of lines separated by `\n`, but the `write` expected an `str`. I feel that its best for `read` and `write` be inverse of each other, such that `write(read())` would write the thing that was read.

For example, I wanted to implement a logic during container restart that the files would only be written whenever the config has changed. In this case, the pseudocode would be like the following if read() and write() are inverse of each other:

```python
existing = file.read()
new_content: str
if existing != new_content:
    file.write(new_content)
```

However, if `read` returns a list and write expects a str, we would need to do a conversion `existing = "\n".join(existing)` on the fly whenever we want to compare whats in the file vs what is expected, which I don't really like. I'd rather check for the FileNotFound exception to be more explicit at the cost of extra try-catch.

Batalex: Fair enough, this is a very valid reason

src/utils/service.py
```diff
@@ -0,0 +1,216 @@
+"""Utility module containing logic of creation and deletion of managed Kyuubi K8s service."""
```
Batalex, line 1: **suggestion**:  What about making this a manager to fit the pattern a bit more? I try (though unsuccessfully sometimes!) to avoid `utils`, which tends to evolve into a dumping ground for unrelated stuff

theoctober19th: I agree. My initial thoughts of making this an util came from an assumption that `utils.service` will be moved to `data-platform-helpers` and us having to change only `from utils.service` to `from data-platform-helpers.service` and the rest being exactly the same.

src/utils/service.py
theoctober19th: Nested function used here to break from nested loop. An inline solution would need:

```python
result = None

for ...1stloop...:
    out = False
    for ...2ndloop...:
        if something:
            result = something
            out = True
            break
        if out:
            break
```

which doesn't please my eyes much.

Instead, the following looks pretty:

```python
def get_something():
    for ...1stloop...:
        for ...2ndloop...:
            if something:
                return something
                
 result = get_something()
```

tests/unit/test_charm.py
theoctober19th: That's a good idea! Updated.

src/utils/service.py
theoctober19th: I've refactored to put this logic in a manager instead of a util.

src/events/base.py
```diff
@@ -68,6 +70,13 @@ def get_app_status(  # noqa: C901
         if self.charm.app.planned_units() > 1 and not self.context.zookeeper:
             return Status.MISSING_ZOOKEEPER.value
 
+        if self.charm.config.get("expose-external", "false") not in VALID_EXPOSE_EXTERNAL_VALUES:
+            return Status.INVALID_EXPOSE_EXTERNAL.value
```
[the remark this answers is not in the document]
theoctober19th: I've removed this validation from here, and have also implemented `structed_config` along this PR.

One thing I noticed with the `structured_config` and thus subsequently the `data_models` lib is that the lib doesn't set the charm to "Blocked" state by default when the config validation fails. And since the `pydantic`'s `ValidationError` is just raised by the charm thus crashing the charm whenever `charm.config` is accessed, the `try / except` and `set_block_status()` should be done at the earliest level of `charm.py` itself and not in the `compute_status` decorator as we have desired for central place of setting app status.

For the scope of this PR, I have left the validation entirely to `structured_config` and thus the charm will crash when `expose-external` is not one of permitted values. As @welpaolo and me discussed, this would not be much of an issue at the moment, but I'd like to have your opinions as well, @deusebio.

Batalex (approved):

src/events/base.py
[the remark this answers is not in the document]
deusebio: I would indeed put a try_catch logic in the constructors of the charm and if uncaught validation errors occurs, I would set the status into blocked status and return straight away (such that hooks are not even registered and run)

src/core/domain.py
```diff
@@ -37,11 +37,14 @@ class Status(Enum):
     INSUFFICIENT_CLUSTER_PERMISSIONS = BlockedStatus(
         "Insufficient cluster permissions. Try: juju trust --scope=cluster <app-name>"
     )
+    INVALID_EXPOSE_EXTERNAL = BlockedStatus("Invalid config option: expose-external")
```
deusebio, line 40: **suggestion, out of scope** I would put a "general" status and message for when config option validation fails. Otherwise, as we increase the number of config options also the statuses would increase. I believe the pydantic messages are fairly nice, and we could just streamline the validation error that we get from pydantic. 

I understand that this may require to rework the use of the enum we have here, and it is more complicated that just one line change. So I would say that it is out of scope (and the solution here is good enought for now), but maybe we should have a thought on how we can change it in the future to make it more scalable.

deusebio (approved):
Looks very good! The introduction of `structured_config` also improved the code quality and cleaness a bit, well-segregating the validation business logic. I had a thought but it is non-blocking, as it would also require a bit of thinking and possibly refactoring. 

Not the refactoring, but maybe the thinking can already be started  :D

Thanks!

src/core/domain.py
theoctober19th: @deusebio I agree with your point.

This line is a leftover after I have removed explicit validation of `expose-external` from our charm.

One thing I noticed with the structured_config and thus subsequently the data_models lib is that the lib doesn't set the charm to "Blocked" state automatically when the config validation fails. And since the pydantic's ValidationError is just raised by the charm thus crashing the charm whenever charm.config is accessed, the try / except and set_block_status() should be done at the earliest level of charm.py itself and not in the compute_status decorator as we have desired for central place of setting app status.

For the scope of this PR, I have left the validation entirely to structured_config and thus the charm will crash when expose-external is not one of permitted values. It would be great if its possible to refactor the `data_models` library a little bit so that it can set a generic "BlockedStatus" by itself when `pydantic` raises a `ValidationError`.

src/events/base.py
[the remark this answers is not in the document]
theoctober19th: Indeed adding be a `try..except` logic in the charm constructor is the most straightforward workaround. However, this will be against our preference of having a single central place of setting charm statust the es. Even if we go along with this as a temporary solution, I'd prefer we look into a more robust solution if possible inside the `data_models` library itself, such that it can set the charm in BlockedState when a ValidationError occurs.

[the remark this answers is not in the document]
theoctober19th: > I think this might cause unexpected issues. By letting the charm error, we make sure that juju will queue events before trying again, whereas an early return before registering events hook handlers would drop events fired while the configuration was wrong.
Maybe sys.exit(1) would work instead?

@Batalex I too experienced this exact issue. `sys.exit(1)` would cause the charm to error. However, exiting from the constructor itself would effectively drop all events fired when the configuration is wrong. The app status was set to desired blocked state, but the unit always had status "error" with message `hook failed: <hook-name>`. In addition to this, the charm also repeatedly tries to resolve the error (see screenshot below). I'm not sure this is what we want. cc: @deusebio @welpaolo 

image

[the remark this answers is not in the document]
theoctober19th: In addition to this, once the charm went to error, `juju config kyuubi-k8s expose-external=false` did not bring it back. The logs showed `juju.worker.uniter awaiting error resolution for "config-changed" hook` and I had to do `juju resolve kyuubi-k8s/0` and wait for a long time to have the unit recover.

welpaolo (approved):
LGTM! Great work here!

src/events/base.py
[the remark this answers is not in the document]
welpaolo: No I don't think that this is what we want. The `sys.exit(1)` strategy does not sound really good. I would prefer to go with the same approach that Kafka and Zookeeper is using.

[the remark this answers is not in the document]
deusebio: I generally don't like charms to go into error state, which feels an uncaught and handled exception. Also, when the config is wrong, my expected ux would be to block and ask the user to change the value, not error out IMHO. 

Having said this, that's a very fair point that my naive suggestion of returning logic in the constructor is arguably even worse, since we are skipping events rather than deferring.

Anyhow, let's not address this here, and don't over engineer this. I agree to keep things with validation error that may be raised, which is the simplest solution from the code perspective and has a resolution path (using juju resolve).

I would still try to have a thought how this can be improved. I will try to ask another couple of teams that I know they are using pydantic

theoctober19th:
PR converted to Draft for additional work.

src/managers/kyuubi.py
```diff
@@ -7,39 +7,75 @@
...
-        s3_info: S3ConnectionInfo | None,
-        metastore_db_info: DatabaseConnectionInfo | None,
-        auth_db_info: DatabaseConnectionInfo | None,
-        service_account_info: SparkServiceAccountInfo | None,
-        zookeeper_info: ZookeeperInfo | None,
+        set_s3_none: bool = False,
+        set_metastore_db_none: bool = False,
+        set_auth_db_none: bool = False,
+        set_service_account_none: bool = False,
+        set_zookeeper_none: bool = False,
```
deusebio, line 49: **praise** Yes, I really like the new approach here! It does simplify and make the code much much much leaner! I really didn't like that changing the signature here would have pretty much effect on the entire code base! WELL DONE!

src/core/context.py
```diff
@@ -64,6 +68,12 @@ def _zookeeper_relation(self) -> Relation | None:
         """The zookeeper relation."""
         return self.model.get_relation(ZOOKEEPER_REL)
 
+    @property
+    def _service_manager(self) -> ServiceManager | None:
```
deusebio, line 72: **thought, suggestion** I find this rather weird and strange, since we are injecting business logic (since it is a manger) into the section (`core`) where we are only dealing with the context, and state (parsing databags and creating domain objects). 

This breaks a bit the dependency structure, where ideally core is rather dependency free, managers depends on core, events depends on managers (and therefore cores). Here we have core that depends on managers, therefore the circular dependency. 

I have a couple of possible suggestions/solutions:
1. Move the service managers to manages and refactor the code accordingly
2. Split the parsing logic of the ServiceManager into the parsing part, and the business-logic part (like creating the service, reconciling, etc)

I understand that as the current code stands, it may seem simpler to just have a manager here. But if we start to deviate from the pattern here, we may well do it elsewhere, breaking the separation of concerns that we currently have, and ending up with more entropy and less maintanable codebases IMHO. I'm curious what others think. I have placed the comment into `suggestion` since it would not be my call, but it is rather for you guys to make a call (and eventually @welpaolo's being the tech lead for the spark project to ).

deusebio (commented):
The codebase looks good! I believe the refactoring of the `update` is very very sane and sensible to me. 

I have one (major?) concern where I would like you guys to have a thought, chime in and come back with a decision. I have my ideas, but this is for you to make a call on this since you will be more involved with maintaining the code base

src/managers/service.py
```diff
@@ -0,0 +1,191 @@
...
+    def get_node(self, unit_name: str) -> lightkube.resources.core_v1.Node:
+        """Return the node for the provided unit name."""
+        node_name = self.get_pod(unit_name).spec.nodeName
+        return self.lightkube_agent.get(
+            res=lightkube.resources.core_v1.Node,
+            name=node_name,
+            namespace=self.namespace,
+        )
+
+    def get_service(self) -> typing.Optional[lightkube.resources.core_v1.Service]:
```
Batalex, line 74: **nitpick**: We could tidy up the types a little in this module

```suggestion
    def get_service(self) -> Service | None:
```
or 
```suggestion
    def get_service(self) -> K8sService | None:
```
if we do `from lightkube.resources.core_v1 import Service as K8sService`

The syntax `| None` is not py3.8 compatible, but we already use it elsewhere. We just might increase the py version in `pyproject.toml` instead

src/core/context.py
Batalex: Alternatively, we could make `kyuubi_address` a method with a `service_manager` parameter. IMO it would address some of your concerns:
- We would avoid instantiating a `ServiceManager` in this module, which is indeed a circular dependency
- Having such a parameter for `kyuubi_address` would only add `ServiceManager` as a type annotation, which would not be a runtime circular dependency (by using `typing.TYPE_CHECKING` to conditionally import it).
- Moreover, using this dependency injection pattern would make this specific function easier to test because we would be able to pass our own `ServiceManager` potentially mocked in the unit tests
- It's more of a nitpick, and a feeble argument, but I usually don't like having an external call behind a non-cached property. Properties usually convey some kind of fast, attribute-like access, whereas we can expect a method call to take some time. It's an even weaker argument because I did exactly that for ZK where we have a property masking an API call, but well, it is what it is

welpaolo: I agree here, we need to remove this manager from the context. I will put this logic in the manager. I would like to continue to follow the pattern. @theoctober19th what do you think here?

welpaolo (commented):
Great work Bikalpa! Please address the concern on the context and then the PR is good to go!

src/core/context.py
deusebio: Not sure I like the pattern proposed by @Batalex if I have to be honest. Seems to me a workaround/patch, to avoid having a runtime error, but still having a similar scope cripple and tension between concerns IMHO. I honestly still see the `typing.TYPE_CHECKING` as a python workaround which I would not honestly encourage too much. 

But I agree on his fourth point about using `cached_property` when using an API call. It does not cost much (just the import of funtools decorator), and it well worth the effort and the pattern in general.

theoctober19th: Thanks everyone for the suggestions, this has been fixed in https://github.com/canonical/kyuubi-k8s-operator/pull/46/commits/67194342b9bae2e88676195b6e0f7e93bd642bc0

src/managers/service.py
theoctober19th: I'll fix this on upcoming PRs since I don't want the Python version bump to cause unforeseen problems.

welpaolo (approved):
LGTM

deusebio (approved):
Looks really good! Very well done! I'm super excited about this feature, and also the discussions that the PR has triggered!

Bibekdhkl:
Amazing🌱🚀

PR #47, opened 2024-12-13
[DPE-6005] Add TLS support in Kyuubi

welpaolo:
This PR add the support of TLS in Kyuubi.


actions.yaml
```diff
@@ -23,4 +23,11 @@ pre-upgrade-check:
 
 resume-upgrade:
   description: |
-    Resume a rolling upgrade after asserting successful upgrade of a new revision.
\ No newline at end of file
+    Resume a rolling upgrade after asserting successful upgrade of a new revision.
+
+set-tls-private-key:
+  description: Set the privates key, which will be used for certificate signing requests (CSR). Run for each unit separately.
```
Batalex, line 29: **typo**: 
```suggestion
  description: Set the private key, which will be used for certificate signing requests (CSR). Run for each unit separately.
```

src/config/kyuubi.py
```diff
@@ -8,18 +8,25 @@
...
 
 class KyuubiConfig(WithLogging):
     """Kyuubi Configurations."""
 
     def __init__(
-        self, db_info: Optional[DatabaseConnectionInfo], zookeeper_info: Optional[ZookeeperInfo]
+        self,
+        db_info: Optional[DatabaseConnectionInfo],
+        zookeeper_info: Optional[ZookeeperInfo],
+        tls: Optional[TLSInfo],
```
Batalex, line 23: **nitpick**: Out of scope, but this type annotation made me realize that we are also using Python 3.8 syntax on this charm though we have 3.10 at runtime. Moving on to 3.10 would mean that we would use `TLSInfo | None`

src/config/kyuubi.py
```diff
@@ -65,9 +72,24 @@ def _ha_conf(self) -> dict[str, str]:
             "kyuubi.ha.zookeeper.auth.digest": self._get_zookeeper_auth_digest(),
         }
 
+    @property
+    def _tls_conf(self) -> dict[str, str]:
```
Batalex, line 76: **nitpick**: Related to previous comment, this type annotation is incompatible with 3.8. No more typing nitpicks, promise!

src/config/kyuubi.py
```diff
@@ -65,9 +72,24 @@ def _ha_conf(self) -> dict[str, str]:
...
+            # enable thrift http frontend with certificate
+            "kyuubi.frontend.thrift.http.ssl.keystore.password": self.tls.keystore_password,
+            "kyuubi.frontend.thrift.http.ssl.keystore.path": self.workload.paths.keystore,
+            "kyuubi.frontend.thrift.http.use.SSL": "true",
+        }
+
     def to_dict(self) -> dict[str, str]:
         """Return the dict representation of the configuration file."""
-        return self._auth_conf | self._ha_conf
+        return self._auth_conf | self._ha_conf | self._tls_conf
```
Batalex, line 92: **praise**: Loving this. So clear and expressive

src/core/workload/__init__.py
```diff
@@ -5,19 +5,83 @@
 """Implementation and blue-print for Kyuubi workloads."""
 
 from abc import abstractmethod
+from pathlib import Path
 
 from common.workload import AbstractWorkload
 
 
+class KyuubiPaths:
+    """Object to store common paths for Kafka."""
```
Batalex, line 14: **typo**:
```suggestion
    """Object to store common paths for Kyuubi."""
```

src/events/kyuubi.py
```diff
@@ -94,3 +107,54 @@ def _on_peer_relation_departed(self, event: ops.RelationDepartedEvent):
         This is necessary for updating status of all units upon scaling up/down.
         """
         self.logger.info("Kyuubi peer relation departed...")
+
+    @compute_status
+    def _on_secret_changed(self, event: SecretChangedEvent) -> None:
```
Batalex, line 112: **question**: Am I missing something, or are we not "observing" the event triggering this handler?

Batalex (changes requested):
Some early feedback, I have yet to review all the files, but I need to drop off for now.

Great job anyway!

src/config/kyuubi.py
deusebio: yes, that's a good point. Can we put a ticket in the backlog for this and link it to this comment?

src/constants.py
```diff
@@ -47,3 +48,14 @@
         "version": "1.9",
     },
 }
+
+SECRETS_UNIT = [
```
deusebio, line 52: **suggestion** This seems a bit too low-level for the constants.py module, and indeed it is used only in `core.context`. I would probably put these constants directly there. 

One of the reason why I'm saying this is also that I haven't been able to place very well `constants.py` in the dependency hierarchy yet. Some of the values here seems to be general, and others seems to be product and substrate specific.

This would call for some thoughts I believe which are out of scope here, but in general I would avoid to place here too much content.

src/core/domain.py
```diff
@@ -258,3 +265,174 @@ def zookeeper_connected(self) -> bool:
...
+
+    def __init__(
+        self,
+        relation: Relation | None,
+        data_interface: Data,
+        component: Unit,
+    ):
+        super().__init__(relation, data_interface, component)
+        self.unit = component
+        self.k8s = ServiceManager(
```
deusebio, line 297: **todo** This is a similar scope cripple as in the PR from @theoctober19th . I suppose we should figure out what to do in the other PR and then rebase and refactor accordingly.

src/core/workload/__init__.py
```diff
@@ -5,19 +5,83 @@
...
 class KyuubiWorkloadBase(AbstractWorkload):
     """Base interface for common workload operations."""
 
-    SPARK_PROPERTIES_FILE = "/etc/spark8t/conf/spark-defaults.conf"
-    HIVE_CONFIGURATION_FILE = "/etc/spark8t/conf/hive-site.xml"
-    KYUUBI_CONFIGURATION_FILE = "/opt/kyuubi/conf/kyuubi-defaults.conf"
-    KYUUBI_ROOT = "/opt/kyuubi"
-    KYUUBI_LOGS = KYUUBI_ROOT + "/logs"
-    KYUUBI_VERSION_FILE = KYUUBI_ROOT + "/RELEASE"
+    paths: KyuubiPaths
```
deusebio, line 84: **praise** I like this pattern which IIRC it is also in Kafka. I believe this makes things more structured and type-safe

src/core/workload/kyuubi.py
```diff
@@ -17,16 +19,26 @@
...
 
+KYUUBI_CONF_PATH = "/opt/kyuubi/conf"
+SPARK_CONF_PATH = "/etc/spark8t/conf"
+KYUUBI_ROOT = "/opt/kyuubi"
+
 
 class KyuubiWorkload(KyuubiWorkloadBase, K8sWorkload, WithLogging):
     """Class representing workload implementation for Kyuubi on K8s."""
 
     def __init__(self, container: Container, user: User = User()):
```
deusebio, line 33: **out-of-scope** not for this PR, but I found it kinda weird that we have something that is substrate specific (the container) in the `core` module, which should be more general. 

Probably in Kyuubi does not matter too much since we only support K8s, so it is not too critical, but coming here from other projects, it got me a bit confused at first. Having a more common pattern would help people find the things straight away, even if they don't know Kyuubi. Anyhow, non-blocking and up to you to decide whether it would be worth it to follow-up (evaluating tradeoff between overenginnering and a strucutre that is more compliant and similar to other charms)

src/events/kyuubi.py
deusebio: I would have the same question actually... :D

src/events/tls.py
```diff
@@ -0,0 +1,208 @@
...
+from core.context import Context
+from core.workload.kyuubi import KyuubiWorkload
+from events.base import BaseEventHandler, compute_status, defer_when_not_ready
+from managers.kyuubi import KyuubiManager
+from managers.tls import TLSManager
+from utils.logging import WithLogging
+
+# from literals import Status
+
+if TYPE_CHECKING:
```
deusebio, line 31: **niptick** I suppose we could just skip this :D

src/managers/tls.py
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env python3
```
deusebio, line 1: **question** I really have a **big question** here... How much have you copied and pasted from Kafka/ZooKeeper here and in `events`?

If we have copied and pasted 100%, that would make my day. If not, also for my understanding, I believe it would be very very beneficial to segregate the general bits and the customizations needed for Kyuubi. I would then put the general class in `src/common` and here we import from common and add the custom bits. The reason why I would do this is because in this way, ideally we would already have a "proposed" implementation for the DP helpers without blocking the work here. As soon as the proposed implementation get accepted in the DP helpers, we can then remove the module in common, and replace here the import from `src.common` to `dataplatformhelpers` . Of course this is *out-of-scope* and completely optional for this PR (unless the split is a no-brainer and easy to achieve), but I wanted to start to float this idea, get your thoughts and possibly implement the split in a future PR

src/events/tls.py
```diff
@@ -0,0 +1,208 @@
+#!/usr/bin/env python3
```
deusebio, line 1: See the comment in `mangers/tls` that refers to this module too

tests/integration/test_tls.py
```diff
@@ -0,0 +1,383 @@
...
+        f"openssl s_client -showcerts -connect {host}:10009 < /dev/null",
+        stderr=subprocess.PIPE,
+        shell=True,
+        universal_newlines=True,
+    )
+
+    assert "TLSv1.3" in response
+    assert "CN = kyuubi" in response
+
+    # get issued certificates
```
deusebio, line 254: **out-of-scope,todo** eventually when we will have a proper relation for retrieving credentials/endpoints, say `kyuubi <> data-integrator` we should probably add the certificate/CA via the relation data, and fetch it using the `data-integrator` action. 

To not forget this, maybe can we add a reminder to this with a code comment with a TODO and/or also adding the link of this comment to the ticket that we have about adding the kyuubi client relation? I believe it is [DPE-3942](https://warthogs.atlassian.net/browse/DPE-3942)

deusebio (commented):
Great Job @welpaolo !!! The changes are very sensible and neat. I believe a bit of a refactoring will be needed because of the K8sManager/ServiceManger refactoring, that we are discussing on #46 

I have some comments, most of them more thoughts and suggestions, but maybe we can already also start to address them when then refactoring for the rebase. 

I'll be happy to re-review after the rebase :D

tests/integration/test_tls.py
```diff
@@ -0,0 +1,383 @@
...
+    bucket_name = s3_bucket_and_creds["bucket"]
+
+    logger.info("Setting up s3 credentials in s3-integrator charm")
+    s3_integrator_unit = ops_test.model.applications[charm_versions.s3.application_name].units[0]
+    action = await s3_integrator_unit.run_action(
+        action_name="sync-s3-credentials", **{"access-key": access_key, "secret-key": secret_key}
+    )
+    await action.wait()
+
+    logger.info("Waiting for s3-integrator app to be idle and active...")
```
theoctober19th, line 114: We'll have this setup done by a helper function once we get the expose-external PR merged. :partying_face:

tests/integration/test_tls.py
```diff
@@ -0,0 +1,383 @@
...
+    jdbc_endpoint_ssl = (
+        jdbc_endpoint
+        + f";ssl=true;trustStorePassword={TRUSTSTORE_PASSWORD};sslTrustStore={TRUSTSTORE_LOCATION}"
+    )
+    logger.info(f"JDBC endpoint with SSL: {jdbc_endpoint_ssl}")
+
+    logger.info(
+        "Testing JDBC endpoint by connecting with beeline" " and executing a few SQL queries..."
+    )
+    process = subprocess.run(
```
theoctober19th, line 333: suggestion: We can already start using the new `run_sql_test_against_jdbc_endpoint(ops_test, test_pod)` helper function here.

theoctober19th (commented):
Great work here! I have a few things here and there, and I'd like to go through another review once we rebase this with the recent changes done for Kyuubi external access.

actions.yaml
```diff
@@ -23,4 +23,11 @@ pre-upgrade-check:
...
+  params:
+    internal-key:
+      type: string
+      description: The content of private key for internal communications with clients. Content will be auto-generated if this option is not specified.
```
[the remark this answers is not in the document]
welpaolo: There are plans to do so but I will keep it out of this PR and add a new ticket for doing so.

src/managers/kyuubi.py
```diff
@@ -30,16 +31,19 @@ def update(
...
+        tls: TLSInfo | None,
     ):
         """Update Kyuubi service and restart it."""
         spark_config = SparkConfig(
             s3_info=s3_info, service_account_info=service_account_info
         ).contents
         hive_config = HiveConfig(db_info=metastore_db_info).contents
-        kyuubi_config = KyuubiConfig(db_info=auth_db_info, zookeeper_info=zookeeper_info).contents
+        kyuubi_config = KyuubiConfig(
+            db_info=auth_db_info, zookeeper_info=zookeeper_info, tls=tls, workload=self.workload
```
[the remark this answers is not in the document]
welpaolo: Done

tests/integration/test_tls.py
welpaolo: Done

src/managers/tls.py
welpaolo: Added [ticket](https://warthogs.atlassian.net/browse/DPE-6398) for next pulse to address

src/core/domain.py
```diff
@@ -259,3 +266,178 @@ def zookeeper_connected(self) -> bool:
...
+    @property
+    def certificate(self) -> str:
+        """The certificate contents for the unit to use for TLS."""
+        return self.relation_data.get("certificate", "")
+
+    @property
+    def ca(self) -> str:
+        """The root CA contents for the unit to use for TLS."""
+        # Backwards compatibility
+        # TODO (zkclient): Remove this property and replace by "" in self.ca_cert
```
theoctober19th, line 361: Do we need this backward compatibility? It seems to me that this is a leftover from Zookeeper code.

theoctober19th (commented):
The PR overall looks fine to me, great work Paolo!

I only have a concern regarding the code related to external access -- has this been tested and validated with the AKS cluster? I'm asking this because earlier during the development of external exposure feature, I had to fall back from deleting and recreating the service to applying the patch over the service, because the former method was failing for the AKS cluster. Also, testing for the actual socket connection with the Kyuubi endpoint proved out to be not much fruitful compared to the the efforts at that time, since there was a lot of over-engineering necessary just to make it work (eg, retry logic, the charm being stuck in blocked state for a long period of time. Is there a specific reason (maybe related to TLS) to revert back and follow this approach instead?

deusebio (approved):
Well done @welpaolo !! Beside the implementation that looks very solid and strutured, you definitely deserve a praise for the tests. I have really liked the TLS integration tests, testing a few use-cases, from TLS working, to certificate renewal and loadbalancer integration. Extremely well done!

src/managers/service.py
```diff
@@ -189,3 +189,36 @@ def reconcile_services(self, expose_external: str):
         self.create_service(
             service_type=desired_service_type, owner_references=pod0.metadata.ownerReferences
         )
+
+    def get_node_ip(self, pod_name: str) -> str:
```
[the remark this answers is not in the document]
welpaolo: This function add return a single IP while the one you mentioned is return a set of IPs

src/managers/service.py
```diff
@@ -189,3 +189,36 @@ def reconcile_services(self, expose_external: str):
...
+            if e.status.code == 403:
+                return ""
+        if not node.status or not node.status.addresses:
+            raise Exception(f"No status found for {node}")
+        for addresses in node.status.addresses:
+            if addresses.type in ["ExternalIP", "InternalIP", "Hostname"]:
+                return addresses.address
+        return ""
+
+    def get_loadbalancer(self) -> str:
```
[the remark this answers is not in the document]
welpaolo: Done

src/managers/service.py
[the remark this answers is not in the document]
theoctober19th: My apologies for the misleading return signature -- I guess I missed to update it to `str` instead of `set[str]`, but the `_get_node_host` returns a single IP (this got changed when we decided a while back that we only return the node of the leader unit when service type is NodePort)

PR #48, opened 2025-01-07
[DPE-6266] Prepare charm for charmcraft 3

Batalex:
## Changes

- Remove support for python < 3.10
- Add support for poetry 2
- Add support for charmcraft 3 based shared workflows
- Upgrade tooling
- Migrate unit tests to scenario v7

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -1,12 +1,12 @@
 # Copyright 2024 Canonical Limited
 # See LICENSE file for licensing details.
 
+from dataclasses import replace
 from unittest.mock import patch
 
 import pytest
 from ops import pebble
-from scenario import Container, Context, Model, Mount, Relation
-from scenario.state import next_relation_id
+from ops.testing import Container, Context, Model, Mount, Relation
 
 from charm import KyuubiCharm
 from constants import (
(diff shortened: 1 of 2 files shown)


marcoppenheimer (approved):

tests/unit/conftest.py
```diff
@@ -53,28 +53,30 @@ def kyuubi_container(tmp_path):
...
-    return Relation(
+    relation = Relation(
         endpoint=S3_INTEGRATOR_REL,
         interface="s3",
         remote_app_name="s3-integrator",
-        relation_id=relation_id,
+    )
+    relation_id = relation.id
+
+    return replace(
```
deusebio, line 78: **praise** TIL!

tests/unit/test_charm.py
```diff
@@ -544,15 +553,15 @@ def test_spark_property_priorities(
     mock_valid_ns,
     mock_s3_verify,
     tmp_path,
-    kyuubi_context,
-    kyuubi_container,
-    s3_relation,
+    kyuubi_context: Context,
+    kyuubi_container: Container,
+    s3_relation: Relation,
```
deusebio, line 558: **out-of-scope, praise** That's very good to add these typings. maybe we should add the static typechecking linting to nudge ourselves more and more towards this?

Batalex: That's definitely something I would be interested in adding to our codebases. `pyright` is now set up in the `pyproject.toml` file (though it's only as a language server right now), the next logical step would be to add it to the linters as a type checker. I did not want to do it in this PR though because I expect to keep the changes to the actual code base to a minimum

PR #56, opened 2025-01-31
[DPE-6477] Add kyuubi-client relation to data-integrator

theoctober19th:
This PR adds support for `kyuubi-k8s` charm to be integrated with `data-integrator` charm.

The tests for the `data-integrator` relation has been pushed via another PR here: https://github.com/canonical/data-integrator/pull/113.

This PR needs to be merged and the changes released for it to be reflected on the other PR.


src/events/provider.py
```diff
@@ -75,21 +75,37 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
...
 
             # Set the JDBC endpoint.
             self.database_provides.set_endpoints(
                 event.relation.id,
-                endpoint,
+                kyuubi_endpoint,
             )
 
+            # Set the JDBC URI
+            self.database_provides.set_uris(event.relation.id, jdbc_uri)
```
theoctober19th, line 90: Here I have proceeded with setting `hostname:port` combination as the "endpoint" and the `jdbc:hive2://hostname:port/` as the "uri" being consistent with other database charms.

Batalex: I am a little confused by this because I thought that we were all migrating from `uris` to `endpoints`, not keeping the two of them

Batalex (approved):
LGTM, my question is not blocking

src/events/provider.py
```diff
@@ -11,38 +11,38 @@
...
+    """Defines event hooks for the provider side of the 'kyuubi-client' relation.
 
     Hook events observed:
         - database-requested
         - relation-broken
     """
 
-    def __init__(self, charm: TypedCharmBase, relation_name: str) -> None:
-        """Constructor for KyuubiClientProvider object.
+    def __init__(self, charm: TypedCharmBase, context: Context, workload: KyuubiWorkloadBase):
```
deusebio, line 35: **praise** well done, I very much prefer the new signature of the constructor

src/events/provider.py
deusebio: I would have preferred to keep the `uri`. The URI seems better than endpoint since it also has the protocol there. Is there a spec about this? I'm not sure I recall a discussion about this, and I'm not aware of the idea of dropping URI

@delgod what's your take on this?

deusebio (approved):
Looks good to me. Let's just resolve the point raised by @Batalex to make sure we align with the ecosystem

src/events/provider.py
theoctober19th: I found [this document](https://docs.google.com/document/d/1UpIZbBz-af2-uNLVTijdbwuW9ASy4IOE5tU9dVEY1Ho/edit?usp=sharing) which mentions the URI (though not the endpoint). I included the endpoint as well as URI here as well because for clients like beeline, its easier to connect with full connection string provided in URI, whereas for clients like `pyhive`, it would be easier extracting hostname:port from the endpoints than a rather convoluted URI.

welpaolo (approved):
LGMT, let's only figure out the endpoint/uri debate :)

src/events/provider.py
Batalex: To give more context, my initial confusion comes from the keys we put in the databags themselves, where we changed `uris` to `endpoints` (see [kafka](https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/kafka_client/v0#example), [mysql](https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/mysql_client/v0#example) or [mongodb](https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/mongodb_client/v0#example).
But since this line is just an internal implementation detail, maybe it does not matter

delgod: `endpoints` is mandatory, and `uris` is optional.
endpoints have strict format `host1,host2` or `host1:port1,host2:port2`.
and `uris` is something which can be passed to the driver, as it is, without additional parsing or modification.

theoctober19th: Thank you Mykola. The PR in its current state is good to be merged, then. Merging it now :)

PR #65, opened 2025-03-03
[DPE-6692] Prevent running actions on non-leader units

Batalex:
## Changes

- Added a check so that running "get-password" on a non-leader unit gives a more understandable message. Since two out of three actions requires running on the leader unit, I also applied the same standard to "get-jdbc-endpoint". From a UX perspective, I think this consistency is nice to have. However, I can easily revert if needed
- Improved type hints a little bit so that we avoid a few `type: ignore` comments

### DIFF
--- a/tests/unit/test_actions.py
+++ b/tests/unit/test_actions.py
@@ -1,3 +1,7 @@
+#!/usr/bin/env python3
+# Copyright 2025 Canonical Limited
+# See LICENSE file for licensing details.
+
 import dataclasses
 import json
 import logging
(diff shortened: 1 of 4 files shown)


theoctober19th (approved):

tests/unit/test_actions.py
```diff
@@ -0,0 +1,130 @@
+import dataclasses
```
welpaolo, line 1: Can you please add the copyright ?

Batalex: Added in 242884c4150cb43cdda7eb84fdcfe78ec7b43016

PR #66, opened 2025-03-05
[DPE-6734] Implement requirer side of Spark Service Account relation

theoctober19th:
This PR implements the requirer side of `spark_service_account` relation. The lib `spark_sa` is currently here in this repo itself. I'll open a separate PR adding this to the `data_platform_libs` once this PR is approved.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -62,7 +62,7 @@ jobs:
           - integration-trust
           - integration-ha
           - integration-external-access
-          - integration-upgrade
+          # - integration-upgrade
           - integration-tls
           - integration-dynamic-allocation
     name: ${{ matrix.tox-environments }}
--- a/src/common/relation/spark_sa.py
+++ b/src/common/relation/spark_sa.py
@@ -23,6 +23,7 @@ from charms.data_platform_libs.v0.data_interfaces import (
     SECRET_GROUPS,
     EventHandlers,
     ProviderData,
+    RelationEventWithSecret,
     RequirerData,
     RequirerEventHandlers,
 )
@@ -130,6 +131,10 @@ class ServiceAccountGrantedEvent(ServiceAccountEvent):
 
 class ServiceAccountGoneEvent(RelationEvent):
     """Event emitted when service account are removed from this relation."""
+
+
+class ServiceAccountPropertyChangedEvent(RelationEventWithSecret):
+    """Event emitted when Spark properties for the service account are changed in this relation."""
 
 
 class SparkServiceAccountRequirerEvents(ObjectEvents):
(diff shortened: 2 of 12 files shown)


.github/workflows/ci.yaml
```diff
@@ -62,7 +62,7 @@ jobs:
           - integration-trust
           - integration-ha
           - integration-external-access
-          - integration-upgrade
+          # - integration-upgrade
```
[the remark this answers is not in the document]
deusebio: yes, that's fine. This is expected that upgrades would not work at this point

src/common/relation/spark_sa.py
```diff
@@ -267,68 +274,81 @@ def __init__(self, charm: CharmBase, relation_data: IntegrationHubRequirerData)
...
+
+        if relation.app == self.charm.app:
+            logging.info("Secret changed event ignored for Secret Owner")
+
+        remote_unit = None
+        for unit in relation.units:
+            if unit.app != self.charm.app:
+                remote_unit = unit
+
+        getattr(self.on, "properties_changed").emit(relation, app=relation.app, unit=remote_unit)
```
deusebio, line 310: **suggestion** I don't know if this would complicate the business logic here, but I'm wondering if we could just make sure the properties have indeed changed before emitting the event. If it is not very critical as in the worst case scenario we would just re-triggering a few more restarts, but I remember in Kafka (maybe @Batalex or @marcoppenheimer remember this too) we have had an issue that this event was retriggered just when the revision was updated but the secret had the same content. 

So I'm wondering whether some safety net would be useful here

deusebio (approved):
Pretty solid work @theoctober19th ! I don't have concern with the code base here which seems very sensible and well structured. I have just a couple of minor, non blocking points

src/common/relation/spark_sa.py
theoctober19th: This check exists both on provider and requirer sides but not yet in the lib. Do you feel that I should migrate this logic to the lib as well?

In the provider side, the new spark-properties are written to relation data if and only if they differ from what is already there in the relation data (see [here](https://github.com/canonical/spark-integration-hub-k8s-operator/blob/sa-relation/src/managers/integration_hub.py#L216)). In the requirer side, the Kyuubi charm restarts itself if and only if the newly computed set of configurations differ from what is already there inside the container (see [here](https://github.com/canonical/kyuubi-k8s-operator/blob/main/src/managers/kyuubi.py#L60)).

I would say that since we have these checks on both sides, we already prevent this. But if you think that we should have this check on the lib-level as well, I'm happy to do so.

Batalex (approved):
LGTM. I have one concern about deferring the service account event if the workload is not ready, but since I never encountered this issue in my numerous deployments, let's keep that in the back of our minds for now

src/common/relation/spark_sa.py
Batalex: We implemented the fix in the dp-libs themselves + for newer ops + juju the fix is already there: https://github.com/canonical/operator/blob/3dd5ace1f9a5ae93ad4174fdde8b4cec9bf350c3/ops/model.py#L1507

src/events/integration_hub.py
```diff
@@ -53,6 +52,12 @@ def _on_account_granted(self, _: ServiceAccountGrantedEvent):
         self.logger.info("Service account received")
         self.kyuubi.update()
```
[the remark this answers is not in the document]
theoctober19th: This is handled by the `@defer_when_not_ready` decorator on the top of the function.

src/common/relation/spark_sa.py
theoctober19th: Super awesome! Thanks for the link, Alex.

PR #67, opened 2025-03-20
[DPE-6884] Enable EKS hostname-based loadbalancer

Batalex:
This PR enables EKS hostname-based loadbalancer services, where we get something alike `ab83aa428bb01464693a33a5215281ea-1597211308.eu-west-1.elb.amazonaws.com` and not a static IP.

We need to change the subject in the CSR; otherwise we get the following error if we try to verify the cert
```
thrift.transport.TTransport.TTransportException: Hostname we connected to "ab83aa428bb01464693a33a5215281ea-1597211308.eu-west-1.elb.amazonaws.com" doesn't match certificate provided commonName "kyuubi-1"
```

We probably have the same issue when it comes to nodeports. This PR does not address this use case, though we can solve it using a similar setup.

## How to test/hack around locally using microk8s?

We can edit `/etc/hosts` to add an entry to something like "mykyuubi.local", pointing to the loadbalancer IP. We then ssh into the charm to edit `events/tls.py` to set the subject name to "mykyuubi.local" and re-relate with self-signed-certificates.
We can then run the impyla test from the spark bundle repository with ca_cert verification without any issue

### DIFF
--- a/src/events/tls.py
+++ b/src/events/tls.py
@@ -4,9 +4,12 @@
 
 """Event handler for related applications on the `certificates` relation interface."""
 
+from __future__ import annotations
+
 import base64
 import os
 import re
+from typing import TYPE_CHECKING
 
 from charms.tls_certificates_interface.v3.tls_certificates import (
     CertificateAvailableEvent,
@@ -14,7 +17,6 @@ from charms.tls_certificates_interface.v3.tls_certificates import (
     generate_csr,
     generate_private_key,
 )
-from ops import CharmBase
 from ops.charm import ActionEvent, RelationBrokenEvent, RelationCreatedEvent, RelationJoinedEvent
 from ops.framework import EventBase
 
@@ -25,11 +27,14 @@ from managers.kyuubi import KyuubiManager
 from managers.tls import TLSManager
 from utils.logging import WithLogging
 
+if TYPE_CHECKING:
+    from charm import KyuubiCharm
+
 
 class TLSEvents(BaseEventHandler, WithLogging):
     """Event handlers for related applications on the `certificates` relation interface."""
 
-    def __init__(self, charm: CharmBase, context: Context, workload: KyuubiWorkload):
+    def __init__(self, charm: KyuubiCharm, context: Context, workload: KyuubiWorkload):
         super().__init__(charm, "tls")
         self.charm = charm
         self.context = context
--- a/src/events/tls.py
+++ b/src/events/tls.py
@@ -89,7 +89,7 @@ class TLSEvents(BaseEventHandler, WithLogging):
                 or self.workload.generate_password(),  # type: ignore
             }
         )
-        subject = os.uname()[1]
+        subject = self.tls_manager.get_subject()
         sans = self.tls_manager.build_sans()
 
         self.logger.info(f"ip: {sans.sans_ip} tls: {sans.sans_dns}")
--- a/src/managers/service.py
+++ b/src/managers/service.py
@@ -105,19 +105,19 @@ class ServiceManager(WithLogging):
         }[expose_external]
 
         match service:
-            case Service(spec=ServiceSpec(type)) if type != expected_service_type:
+            case Service(spec=ServiceSpec(type=type)) if type != expected_service_type.value:
                 return ""
-            case Service(spec=ServiceSpec(type=_ServiceType.CLUSTER_IP)):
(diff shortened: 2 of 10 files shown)


src/managers/service.py
```diff
@@ -3,14 +3,29 @@
...
+    ServiceStatus,
+)
+from lightkube.models.meta_v1 import ObjectMeta
+from lightkube.resources.core_v1 import Node, Pod, Service
 
 from constants import JDBC_PORT
 from utils.logging import WithLogging
 
 
+class LbHost(str):
```
deusebio, line 23: **niptick** why not - just for the sake of expressiveness and validation - use a dataclass like:

```
LbHost:
    address: str
    port: int
```

I don't know, my OCD gets a bit hurt to just override the str with no special characterization. With the one above we also have a bit more validation for the integer in the port.

src/managers/service.py
```diff
@@ -62,65 +77,76 @@ def _get_node_address(node) -> str:
...
-        except lightkube.core.exceptions.ApiError as e:
+        except ApiError as e:
             if e.status.code == 404:
                 return None
             raise
 
         return service
 
-    def get_service_endpoint(self, expose_external: str):
+    def get_service_endpoint(self, expose_external: str) -> str:
```
deusebio, line 104: **suggestion** Why not providing a consistent output?

deusebio (approved):
I'm approving as the following is really not a super concerning point, but rather a suggestion. 

My OCD got a bit hurt (you see the comment), then I understood why you did this, to separate the behavior later in the code using `match` statements, which seemed a bit of a workaround and inconsistent signatures, and create some question marks for other developers (like: "I thought this function returned str, not LbHost"). My suggestion would just be to use a proper class, and use a type, as I suggest in the comment (since you are already defining one). 

In this way you can have rather explicit and readable signatures, e.g. 

```
def get_service_endpoint(self, expose_external: str) -> LbHost:
```

and then differentiate the behaviour easily:

```
if loadbalance.type == "IP":
    sans_ip.append(loadbalancer.split(":")[0])
```

using match for this seems a bit beyond its scope, making signature/logic less clear. I would generally use match when classes has similar hierarchy, like:

```
class Animal: ...

class Dog(Animal): ...
class Cat(Animal): ...

match var:
    case Dog(name):
          return f"bark {name}"
    case Cat(name):
         return f"meawaw {name}"
```

but that's not the case here. Anyhow, my 2 cents, but it is your call to make to have a code that you guys are happy to work with

theoctober19th:
I'll review the PR once Enrico's comments are addressed and the PR is ready for review again.

src/managers/service.py
Batalex: Refactored to use a dataclass in 8d8a5e6f8d06917429f648972e2c06a56e890f88

theoctober19th (approved):
LGTM. Well done, Alex!

src/events/tls.py
```diff
@@ -25,11 +27,14 @@
...
 
+if TYPE_CHECKING:
+    from charm import KyuubiCharm
+
 
 class TLSEvents(BaseEventHandler, WithLogging):
     """Event handlers for related applications on the `certificates` relation interface."""
 
-    def __init__(self, charm: CharmBase, context: Context, workload: KyuubiWorkload):
+    def __init__(self, charm: KyuubiCharm, context: Context, workload: KyuubiWorkload):
```
deusebio, line 37: **question** I'm not a bit fan of this pattern. I personally feel it is a bit of a code-smell that hints circular dependency (even if just for the linting part), which I would nudge to avoid, unless it is needed. But I don't understand why this is needed in this context

deusebio (approved):
Looks very good! I have one minor on one thing I'd nudge against, but I would leave it to you guys to make a call on.

src/events/tls.py
Batalex: Yes, that was a fix for a type check violation. `TLSEvents` inherits from `BaseEventHandler` which declares a `charm` attr with the `KyuubiCharm` type. So we cannot go back to a more general `BaseCharm` assignment two lines below 
Full disclosure: according to git blame, I was the one who made this previous change as well in #65 to remove a whole lotta `# type: ignore` instructions.

I am going to go ahead and merge this as-is. My plan with all of those tiny fixes along the way is to have type safety first so that we can enable mypy to prevent any regression, and then refactor what needs to be

theoctober19th: +1 with Alex.

PR #68, opened 2025-04-03
[DPE-6785] Use spark_service_account from charmhub charm-lib

theoctober19th:
Use `spark_service_account` lib from charm libs, instead of a local one.


deusebio (approved):
Looks very good!

Batalex (changes requested):
Let's update the lib, LGTM after that

Batalex (approved):

PR #69, opened 2025-04-04
[DPE-6704] Apache iceberg integration with Kyuubi

theoctober19th:
* Add Iceberg support for Kyuubi
* Add integration tests for Apache Iceberg support
* Initialize Hive Metastore schema in postgres upon relation with postgres
* Add new integration tests for external metastore with Kyuubi
* Refactor tests to split tests into different components


tests/integration/test_observability.py
```diff
@@ -0,0 +1,194 @@
...
+        idle_period=30,
+    )
+
+    # These two relations --though essential to publishing-- are not set.
+    # (May change in the future?)
+    try:
+        await ops_test.model.integrate(
+            f"{COS_AGENT_APP_NAME}:grafana-dashboards-provider", "grafana"
+        )
+    except juju.errors.JujuAPIError:
```
Batalex, line 133: **question**: I know this is a copy/paste from the original test in `test_charm`, but why do we need to silence all those exceptions?

tests/integration/test_relation.py
```diff
@@ -0,0 +1,254 @@
...
+    time.sleep(30)
+
+    # Fetch number of users excluding the default admin user
+    with connection.cursor() as cursor:
+        cursor.execute(""" SELECT username, passwd FROM kyuubi_users WHERE username <> 'admin' """)
+        num_users_after = cursor.rowcount
+
+    connection.close()
+
+    # Assert that a new user had indeed been created
```
Batalex, line 247: **nitpick**: The comment is not super clear as to what the code is doing

Batalex (commented):
Overall LGTM.
It's really nice to have a proper segmentation in the tests; good job!

src/events/base.py
```diff
@@ -58,6 +59,10 @@ def get_app_status(  # noqa: C901
         if not k8s_manager.is_service_account_valid():
             return Status.INVALID_SERVICE_ACCOUNT.value
 
+        metastore_manager = HiveMetastoreManager(self.workload)
+        if self.context.metastore_db and not metastore_manager.is_metastore_valid():
```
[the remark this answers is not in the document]
theoctober19th: Can't say the exact time in seconds, but won't take long since this is just a call to `schematool -validate`. The time would be kinda similar / comparable to other checks we do (for example, check if service account exists, check if namespace exists, check if object storage is valid, etc.

tests/integration/test_relation.py
theoctober19th: Fixed in 047d7e4

tests/integration/test_metastore.py
```diff
@@ -0,0 +1,307 @@
...
+
+    # Assert that Kyuubi charm is in active state again
+    assert check_status(ops_test.model.applications[APP_NAME], Status.ACTIVE.value)
+
+
+@pytest.mark.abort_on_fail
+async def test_read_write_with_valid_schema_metastore_again(ops_test):
+    """Test whether previously written data can be read as well as new data can be written."""
+    assert await validate_sql_queries_with_kyuubi(
+        ops_test=ops_test, query_lines=["USE dbext;", "SELECT * FROM text;"]
```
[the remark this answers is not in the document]
theoctober19th: Fixed in 047d7e4

tests/integration/helpers.py
```diff
@@ -676,3 +677,42 @@ async def fetch_spark_properties(ops_test: OpsTest, unit_name: str) -> dict[str,
         temp_file.seek(0)
         props = PropertyFile.read(temp_file.name).props
         return props
+
+
+async def validate_sql_queries_with_kyuubi(
```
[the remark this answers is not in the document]
theoctober19th: Fixed in 047d7e4.

src/managers/hive_metastore.py
```diff
@@ -0,0 +1,75 @@
...
+        except ops.pebble.ExecError as e:
+            # ExceError is raised when the return code is not 0
+            return 1, e.stdout or "", e.stderr or ""
+        except Exception as e:
+            logger.exception(e)
+            return 1, "", str(e)
+
+    def is_metastore_valid(self) -> bool:
+        """Validate the metastore schema and return if it is valid."""
+        retcode, stdout, stderr = self._run_schematool_command("-validate")
```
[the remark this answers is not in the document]
theoctober19th: Fixed in 047d7e4.

.github/workflows/ci.yaml
```diff
@@ -59,12 +59,17 @@ jobs:
...
+          - integration-observability
+          - integration-tls
           - integration-ha
           - integration-external-access
-          # - integration-upgrade
-          - integration-tls
           - integration-dynamic-allocation
+          - integration-upgrade
+          - integration-iceberg
+          - integration-relation
```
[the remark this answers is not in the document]
theoctober19th: Fixed in 047d7e4.

tests/integration/test_observability.py
theoctober19th: My guess is that these essential relations don't exist in the bundle and are added manually later. If in case in future these relations come from the `cos` bundle itself, we need to be safe that this code doesn't raise exception right away. However, I don't know why we silence these exceptions though, we could have simply logged them.

theoctober19th: I've replaced the `pass` with `logger.error()` calls.

.github/workflows/ci.yaml
```diff
@@ -59,12 +59,17 @@ jobs:
       matrix:
         tox-environments:
           - integration-charm
-          - integration-trust
+          - integration-metastore
+          - integration-auth
+          - integration-observability
+          - integration-tls
```
deusebio, line 65: **praise** well done on breaking down those tests. I believe it is very very sensible

src/constants.py
```diff
@@ -27,13 +27,17 @@
 
 # Literals related to Kyuubi
 JDBC_PORT = 10009
+SPARK_DEFAULT_CATALOG_NAME = "spark_catalog"
```
deusebio, line 30: **question** I remember that we discussed this, but I don't remember the conclusion. Should the SPARK_DEFAULT_CATALOG_NAME be configurable or were there blockers/issues to allow this?

src/managers/hive_metastore.py
```diff
@@ -0,0 +1,75 @@
...
+            )
+            return
+
+        # Attempt initialization with dry run first
+        retcode, stdout, stderr = self._run_schematool_command(
+            "-initSchemaTo", schema_version, dry_run=True
+        )
+        if retcode != 0:
+            logger.error(
+                f"Cannot safely initialize Hive schema in metastore database. stdout={stdout}; stderr={stderr}"
```
deusebio, line 63: **question** Do we want this to surface a bit more than a log? Should the charm go into blocked status if we can't initialize the schema? seems important

deusebio (commented):
Very solid work! I just have a couple of questions that I'd like clarity and understanding whether the failure in the initializing the schema is a bit more important than drowning that in the logs

tests/integration/test_ha.py
```diff
@@ -2,6 +2,8 @@
 # Copyright 2024 Canonical Limited
 # See LICENSE file for licensing details.
 
+# TODO: Revisit this test after recent updates in the purpose of Kyuubi <> Zookeeper relation
```
welpaolo, line 5: What do you mean here ?

src/constants.py
theoctober19th: This is the name of the default spark catalog, which is always `spark_catalog`. I guess what you refer to is the name of the catalog in which Iceberg will be enabled, this is configured via the configuration option `iceberg-catalog-name`, which by default is `iceberg` unless specfied otherwise.

src/managers/hive_metastore.py
theoctober19th: Yes, the charm should go to blocked state, and it does indeed do so. We do this [here](https://github.com/canonical/kyuubi-k8s-operator/pull/69/files#diff-51d68eefb2ecff79297dcb188c7c39df6f5410e643163d6c1402bb0c2c8a2ae2R63) and we also test this behavior with integration test [here](https://github.com/canonical/kyuubi-k8s-operator/pull/69/files#diff-643a25590253704a0be6643d87e9c572ffd911d68aba1e195eecab705f83e75dR256).

welpaolo (approved):
LGTM, I left some comments.

tests/integration/test_ha.py
theoctober19th: Before K8s exposure, HA used to be tied up with Zookeeper because we used Zookeeper for service discovery. This is no longer the case. We use Zookeeper only for engine discovery at the moment, and thus our tests involving zookeeper need to be updated for this particular role of zookeeper. For example, we don't test the "engine discovery" behavior of zookeeper.

metadata.yaml
```diff
@@ -34,7 +34,7 @@ resources:
   kyuubi-image:
     type: oci-image
     description: OCI image for kyuubi
-    upstream-source: ghcr.io/canonical/charmed-spark-kyuubi@sha256:9268d19a6eef91914e874734b320fab64908faf0f7adb8856be809bc60ecd1d0
+    upstream-source: ghcr.io/canonical/charmed-spark-kyuubi@sha256:86fc84c8d01da25f756bebbae17395ef9702a8fd855565a4a80ed5d4f8024708
```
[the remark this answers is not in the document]
theoctober19th: Accepted the commit suggestion. Thanks.

src/managers/hive_metastore.py
deusebio: Oh I see. That's fine then. One suggestion (but at this point it is completely optional) that I have is that you could return meaningful information about of the function, like returning "True" is the initialization was fine, "False" if there were issues with that. And the two events could be handled in the outer function. 

What got me a bit confused here was that the function returned the same output in those two cases.

deusebio (approved):
Thank you for the responses. The code look good and I don't have concerns with this! LGTM!

src/managers/hive_metastore.py
theoctober19th: I chose to do this at the lowest level because the failures could be due to different reasons. For example, it can fail due to the initial validation failing (meaning that the metastore is not in a state where the initialization can be applied), or during the initialization (meaning something went wrong there).

PR #70, opened 2025-04-08
[MISC] Endpoint is not string

dragomirp:


src/events/provider.py
```diff
@@ -106,7 +111,7 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
                     event.relation.id, self.context.unit_server.ca_cert
                 )
 
-        except (Exception) as e:
+        except Exception as e:
             logger.exception(e)
             self.charm.unit.status = BlockedStatus(str(e))
```
dragomirp, line 116: This status seems to be overwritten by other status changes.

src/events/provider.py
```diff
@@ -57,33 +57,38 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
...
             username = f"relation_id_{event.relation.id}"
             password = auth.generate_password()
             auth.create_user(username=username, password=password)
 
             kyuubi_endpoint = service_manager.get_service_endpoint(
                 expose_external=self.charm.config.expose_external
             )
-            jdbc_uri = f"jdbc:hive2://{kyuubi_endpoint}/" if kyuubi_endpoint else ""
+            jdbc_uri = (
+                f"jdbc:hive2://{kyuubi_endpoint.host}:{kyuubi_endpoint.port}/"
```
dragomirp, line 83: Might be easier to add `__str__()` to `IPEndpoint` and `DNSEndpoint` and cast instead.

dragomirp:
Local tests seemed fine, but I'm not familiar with the charm.

src/events/provider.py
Batalex: Yes, absolutely. Added to our housekeeping todo list, thanks

Batalex (approved):
LGTM.

My apologies, as I am the one who introduced this regression with the work on #67. Thanks, for both fixing the issue as well a pointing to us that there is a large gap in our testing coverage

deusebio (approved):
Great catch @dragomirp !! Thank you for not only raising this, but also fixing it! very appreciated!

src/events/provider.py
theoctober19th: That's true, because our status are handled centrally. Added to our list of housekeeping tasks. Thanks!

theoctober19th (approved):
Thanks for the lovely PR @dragomirp <3

PR #72, opened 2025-04-24
[DPE-6890] Wire up loadbalancer-extra-annotations to service creation

Batalex:
# Changes

- Implement `loadbalancer-extra-annotations` to add annotation to the K8s LB service

**Important** 
I quickly checked on the testing AKS playground if we can deal with certs just by using the "correct" annotation setting up the hostname, and it appears that we will need some additional logic if we want to include a hostname in the certs from an annotation.
The current ManSol deployment strategy sets a hostname as ca name at the self-signed-certificates level, but the get-jdbc-endpoint action will return an IP. This IP will be set in the welcome email template. We might need to work with ManSol on that aspect.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -4,20 +4,20 @@
 options:
   namespace:
     type: string
-    description: The namespace to be used by driver to create executor pods. If not configured, the model namespace will be used. 
-    default: ''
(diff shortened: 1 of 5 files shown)


theoctober19th (approved):
LGTM. Thanks Alex!

welpaolo (approved):
LGTM

PR #73, opened 2025-04-24
[DPE-6830] Housekeeping and jubilant migration

Batalex:

PR #74, opened 2025-04-28
[DPE-7061] Make auth db relation mandatory

Batalex:
## Changes

- Make `auth-db` relation mandatory. The charm won't start without it, with an appropriate blocked status
- This PR includes a bunch of housekeeping items from #73.  I had already fixed some stuff there, and was more confident in creating a branch from this point. It should be fairly straightforward to review though.

The integration tests are super verbose now, with the username and password needed everywhere. I kindly ask for your understanding here: since we have a full rewrite to jubilant coming up, I did not bother with making the changes as maintainable as I would usually do.

### DIFF
--- a/src/events/actions.py
+++ b/src/events/actions.py
@@ -6,14 +6,14 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, cast
 
 from ops.charm import ActionEvent
 
 from constants import DEFAULT_ADMIN_USERNAME
 from core.context import Context
-from core.domain import Status
-from core.workload import KyuubiWorkloadBase
+from core.domain import DatabaseConnectionInfo, Status
+from core.workload.kyuubi import KyuubiWorkload
 from events.base import BaseEventHandler
 from managers.auth import AuthenticationManager
 from managers.kyuubi import KyuubiManager
@@ -27,7 +27,7 @@ if TYPE_CHECKING:
 class ActionEvents(BaseEventHandler, WithLogging):
     """Class implementing charm action event hooks."""
 
-    def __init__(self, charm: KyuubiCharm, context: Context, workload: KyuubiWorkloadBase):
+    def __init__(self, charm: KyuubiCharm, context: Context, workload: KyuubiWorkload):
         super().__init__(charm, "action-events")
 
         self.charm = charm
@@ -35,7 +35,6 @@ class ActionEvents(BaseEventHandler, WithLogging):
         self.workload = workload
 
         self.kyuubi = KyuubiManager(self.workload, self.context)
-        self.auth = AuthenticationManager(self.context.auth_db)
         self.service_manager = ServiceManager(
             namespace=self.charm.model.name,
             unit_name=self.charm.unit.name,
@@ -111,7 +110,8 @@ class ActionEvents(BaseEventHandler, WithLogging):
                 event.fail(msg)
                 return
 
-        password = self.auth.get_password(DEFAULT_ADMIN_USERNAME)
+        auth = AuthenticationManager(cast(DatabaseConnectionInfo, self.context.auth_db))
+        password = auth.get_password(DEFAULT_ADMIN_USERNAME)
         event.set_results({"password": password})
 
     def _on_set_password(self, event: ActionEvent) -> None:
@@ -147,16 +147,17 @@ class ActionEvents(BaseEventHandler, WithLogging):
                 event.fail(msg)
                 return
 
-        password = self.auth.generate_password()
+        auth = AuthenticationManager(cast(DatabaseConnectionInfo, self.context.auth_db))
+        password = auth.generate_password()
 
         if "password" in event.params:
             password = event.params["password"]
 
-        if password == self.auth.get_password(DEFAULT_ADMIN_USERNAME):
+        if password == auth.get_password(DEFAULT_ADMIN_USERNAME):
             event.log("The old and new passwords are equal.")
             event.set_results({"password": password})
             return
 
-        self.auth.set_password(DEFAULT_ADMIN_USERNAME, password)
+        auth.set_password(DEFAULT_ADMIN_USERNAME, password)
 
         event.set_results({"password": password})
--- a/src/events/auth.py
+++ b/src/events/auth.py
@@ -85,5 +85,5 @@ class AuthenticationEvents(BaseEventHandler, WithLogging):
         """
         self.logger.info("Authentication database relation departed")
         # TODO: I don't think we should be doing this, is it our responsibility?
-        auth = AuthenticationManager(cast(DatabaseConnectionInfo, self.context.auth_db))
-        auth.remove_auth_db()
+        # auth = AuthenticationManager(cast(DatabaseConnectionInfo, self.context.auth_db))
+        # auth.remove_auth_db()
--- a/src/events/auth.py
+++ b/src/events/auth.py
@@ -4,24 +4,31 @@
 
 """Authentication related event handlers."""
 
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, cast
+
 from charms.data_platform_libs.v0.data_interfaces import (
     DatabaseCreatedEvent,
     DatabaseRequirerEventHandlers,
 )
-from ops import CharmBase
 
 from core.context import Context
-from core.workload import KyuubiWorkloadBase
+from core.domain import DatabaseConnectionInfo
+from core.workload.kyuubi import KyuubiWorkload
 from events.base import BaseEventHandler, compute_status, defer_when_not_ready
 from managers.auth import AuthenticationManager
 from managers.kyuubi import KyuubiManager
 from utils.logging import WithLogging
 
+if TYPE_CHECKING:
+    from charm import KyuubiCharm
+
 
 class AuthenticationEvents(BaseEventHandler, WithLogging):
     """Class implementing PostgreSQL metastore event hooks."""
 
-    def __init__(self, charm: CharmBase, context: Context, workload: KyuubiWorkloadBase):
+    def __init__(self, charm: KyuubiCharm, context: Context, workload: KyuubiWorkload) -> None:
         super().__init__(charm, "s3")
 
         self.charm = charm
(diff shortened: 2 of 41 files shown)


src/events/auth.py
```diff
@@ -67,11 +78,5 @@ def _on_auth_db_relation_removed(self, event) -> None:
...
     def _on_auth_db_relation_departed(self, event) -> None:
-        """Handle the event when the authentication database relation is departed.
-
-        Until this point, the relation data is still there and hence the credentials
-        can be fetched for one last time in order to remove authentication database.
-        """
+        """Handle the event when the authentication database relation is departed."""
         self.logger.info("Authentication database relation departed")
-        auth = AuthenticationManager(self.context.auth_db)
-        auth.remove_auth_db()
```
Batalex, line 77: I had my suspicion that this was an issue in the housekeeping PR and made a note there. Turns out, it was indeed an issue and a blind spot in our test coverage: this event is emitted when we scale down kyuubi, meaning that we would remove the user table and thus break the charm.

welpaolo: Great catch Alex!

src/events/actions.py
```diff
@@ -147,16 +147,17 @@ def _on_set_password(self, event: ActionEvent) -> None:
                 event.fail(msg)
                 return
 
-        password = self.auth.generate_password()
+        auth = AuthenticationManager(cast(DatabaseConnectionInfo, self.context.auth_db))
```
theoctober19th, line 150: **praise**: thanks for moving this down to the event handler. This will solve some race conditions we saw during marketplace testing.

src/events/provider.py
```diff
@@ -122,7 +129,8 @@ def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
         if not self.charm.unit.is_leader():
             return
 
-        auth = AuthenticationManager(self.context.auth_db)
+        # FIXME: There is not guarantee here
```
theoctober19th, line 132: WDYM? This relation (provider) relation is different from Kyuubi <> Postgres(Auth DB) relation -- and we block Kyuubi if it is not related over auth db interface anway.

src/managers/kyuubi.py
```diff
@@ -48,14 +48,24 @@ def update(
...
-    ):
+    ) -> None:
         """Update Kyuubi service and restart it."""
         metastore_db_info = None if set_metastore_db_none else self.context.metastore_db
         auth_db_info = None if set_auth_db_none else self.context.auth_db
         service_account_info = None if set_service_account_none else self.context.service_account
         zookeeper_info = None if set_zookeeper_none else self.context.zookeeper
         tls_info = None if set_tls_none else self.context.tls
 
+        # auth is mandatory
```
theoctober19th, line 59: I think having this check here is a bit of an overkill. For example, we don't do this if Integration Hub hasn't been related, or for example, the object storage is not available. And, we will put the charm in blocked state, anyway.

This check potentially will result in us adding further checks in the future for all other "blocking" checks (eg, integration hub, object storage, etc.) and thus potentially we will have to repeat the validation logic for all blocking conditions at two places in the charm.

tests/integration/conftest.py
```diff
@@ -92,6 +93,16 @@ def charm_versions() -> IntegrationTestsCharms:
                 "trust": True,
             }
         ),
+        auth_db=TestCharm(
+            **{
+                "name": "postgresql-k8s",
+                "channel": "14/stable",
+                "revision": 281,
+                "series": "jammy",
+                "alias": "auth-db",
```
theoctober19th, line 102: **nit**: now that we have 2 postgres instances, I'd also maybe want to change the other app alias to `metastore-db` for clarity.

tests/integration/conftest.py
```diff
@@ -109,7 +120,7 @@ def charm_versions() -> IntegrationTestsCharms:
                 "revision": 70,
                 "series": "jammy",
                 "alias": "zookeeper",
-                "num_units": 3,
+                "num_units": 1,
```
theoctober19th, line 123: +1 on reducing the number of units. I also think we should rethink a little bit on the tests related to zookeeper in the near future now that we have changed the role of zookeeper in our bundle (it exists for service discovery and not for HA, but our current tests revolve around Zookeeper HA)

src/events/auth.py
Batalex: > I think this should probably be done in relation-broken hook instead

It's not the requirer responsibility to cleanup any resource on the provider side, right?

tests/integration/test_auth.py
```diff
@@ -55,76 +47,12 @@ async def test_deploy_minimal_kyuubi_setup(
...
-    assert check_status(ops_test.model.applications[APP_NAME], Status.ACTIVE.value)
-    assert (
-        ops_test.model.applications[charm_versions.integration_hub.application_name].status
-        == "active"
-    )
-    assert ops_test.model.applications[charm_versions.s3.application_name].status == "active"
-
-
-@pytest.mark.abort_on_fail
-async def test_sql_queries_no_authentication(ops_test):
```
theoctober19th, line 73: I think these tests are still relevant here.

tests/integration/test_auth.py
```diff
@@ -199,24 +127,3 @@ async def test_set_password_action(ops_test: OpsTest):
     assert await validate_sql_queries_with_kyuubi(
         ops_test=ops_test, username=username, password=new_password
     )
-
-
-@pytest.mark.abort_on_fail
-async def test_remove_authentication(ops_test: OpsTest, test_pod, charm_versions):
```
theoctober19th, line 205: I think this test is still relevant here. We can assert the charm goes to blocked state in addition as well.

tests/integration/test_charm.py
```diff
@@ -301,8 +352,22 @@ async def test_integration_with_zookeeper(ops_test: OpsTest, charm_versions):
     await ops_test.model.wait_for_idle(
         apps=[APP_NAME, charm_versions.zookeeper.application_name], timeout=1000, status="active"
     )
+    kyuubi_leader = await find_leader_unit(ops_test, app_name=APP_NAME)
+    assert kyuubi_leader is not None
+
+    logger.info("Running action 'get-password' on kyuubi-k8s unit...")
```
theoctober19th, line 358: One thing that I don't like here is the repetition of these lines across all test functions. However, we can defer that for now, since we want to have this merged ASAP. We can clean this later in a different PR, but let's create a marker/ticket in the backlog so we don't miss it later.

tests/integration/test_iceberg.py
```diff
@@ -52,23 +52,28 @@ async def test_deploy_kyuubi_setup(
         status="active",
     )
 
-    # Assert that all charms that were deployed as part of minimal setup are in correct states.
-    assert check_status(ops_test.model.applications[APP_NAME], Status.ACTIVE.value)
```
theoctober19th, line 56: **question**: Why do we remove this? I think assertion that charm is in active state is a nice-to-have.

theoctober19th (approved):
The PR looks good overall.

I have a few comments, however we can have them addressed later in a different PR as well.

src/events/provider.py
Batalex: Kyuubi might be blocked from a status perspective, but there is nothing preventing the following scenario:
- We have a working setup with auth db and a client
- We remove the auth db relation, kyuubi goes into blocked status
- We remove the client relation. There is nothing preventing this event handler from running, so `self.context.auth_db` is None actually

I use `cast` as a way to tell the type checker "in this specific situation, trust me I know what type is my var", but here, it could be none.

Right now I care about getting the green tick, but subsequent PRs should focus on type exactitude. I should check whether `self.context.auth_db` is None or not, and then act on it

tests/integration/test_charm.py
Batalex: Yeah I'll streamline all of these with jubilant

tests/integration/test_iceberg.py
Batalex: It's redundant with the call to `wait_for_idle` just above since it already waits for active state on all three apps

tests/integration/conftest.py
Batalex: You are going to love the next PR then :D We'll have full type safety and autocompletion on those elements

src/managers/kyuubi.py
Batalex: We need to distinguish between "blocked state" and "workflow is not running" IMO. If the charm is in blocked status because of a missing object storage and you try to use it only to get an error, that's on you.
If the charm is in blocked status because of missing auth but you can still interact with it using beeline that's a whole different problem.

Were you thinking of something else? I might have misunderstood

tests/integration/test_ha.py
```diff
@@ -81,7 +79,21 @@ async def test_build_and_deploy_cluster_with_no_zookeeper(
 
 @pytest.mark.abort_on_fail
 async def test_standalone_kyuubi_works_without_zookeeper(ops_test: OpsTest, test_pod):
-    assert await run_sql_test_against_jdbc_endpoint(ops_test, test_pod)
+    kyuubi_leader = await find_leader_unit(ops_test, app_name=APP_NAME)
```
welpaolo, line 82: For a future PR, can we move the password retrial logic in an helper function?

welpaolo (approved):
Great work Alex! I just left a minor comment

src/events/provider.py
theoctober19th: I get your point. Thanks for the clarification.

src/managers/kyuubi.py
theoctober19th: Thanks for the clarification, Alex. I agree with it.

tests/integration/conftest.py
theoctober19th: Awesome :tada:

tests/integration/test_auth.py
theoctober19th: I'd still prefer to keep this here since this test here, while in `test_charm` we can test just the happy path (since we have different modules that intensively test each specific aspect of the charm. However, this is not a big issue, and I leave this upto you.

tests/integration/test_charm.py
theoctober19th: :tada:

tests/integration/test_auth.py
Batalex: Added in f19f344

PR #76, opened 2025-04-29
[DPE-6830] Jubilant migration

Batalex:
# Changes

- Migrate all integrations tests to jubilant. The integration tests are a little faster compared to `pytest-operator`, which is nice since adding mandatory auth everywhere tremendously increased the testing time.
- The integration tests are now type checked. I am especially excited about this because, as you can see, the total diff in lines of code number remains negative by more than a thousand lines despite the boilerplate.


tox.ini
```diff
@@ -97,4 +98,4 @@ pass_env =
     CI
 commands =
     poetry install --with integration
-    poetry run pytest -vv --tb native --log-cli-level=INFO -s {posargs} {[vars]tests_path}/integration/{env:TESTFILE}
+    poetry run pytest -vv --tb native --log-cli-level=INFO -s {posargs} {[vars]tests_path}/integration/{env:TESTFILE} -x
```
Batalex, line 101: **note**: `-x` pytest option is here to somewhat replace the `@abort_on_fail` decorator

tests/integration/helpers.py
```diff
@@ -468,144 +434,148 @@ async def deploy_minimal_kyuubi_setup(
...
     charm_config = {"namespace": namespace, "service-account": username}
+    juju.config(APP_NAME, charm_config)
 
-    await ops_test.model.applications[APP_NAME].set_config(charm_config)
     logger.info("Waiting for kyuubi-k8s app to settle...")
-    await ops_test.model.wait_for_idle(apps=[APP_NAME], status="blocked", idle_period=20)
-    assert check_status(
-        ops_test.model.applications[APP_NAME], Status.MISSING_INTEGRATION_HUB.value
-    )
+    status = juju.wait(jubilant.all_blocked)
```
theoctober19th, line 451: Wow, this makes our life a lot easier! :relieved:

tests/integration/test_charm.py
```diff
@@ -94,395 +68,280 @@ async def test_deploy_s3_integrator(ops_test: OpsTest, charm_versions, s3_bucket
...
-
-    assert len(executor_pod_names) == len(expected_executor_pod_names)
-    assert set(executor_pod_names) == set(expected_executor_pod_names)
+    password = fetch_password(juju)
+    assert validate_sql_queries_with_kyuubi(juju, username=username, password=password)
+
+
+# TODO: rewrite
+# @pytest.mark.skip(reason="This tests need re-write and fixes on integration hub level")
+# def test_read_spark_properties_from_secrets(ops_test: OpsTest):
```
theoctober19th, line 280: In my opinion, we can remove this test altogether because the thing it is trying to test is already obsolete in our case. The Kyuubi charm no longer gets notified of the changes performed at user level using `spark8t` tool, but is concerned with only cluster level configurations provided to it by the Integration Hub. As far as the test that Kyuubi is notified of the cluster level configurations, we already do that in this file. As a result, I think we can drop this function altogether.

theoctober19th (approved):
Wow, that was a really long PR, thank you so much Alex for the migration! The changes look fine to me, left a comment regarding deleting an unused test altogether.

welpaolo (approved):
Great work Alex! Please check if we can update Jubilant to 1.0.1

tests/integration/test_charm.py
Batalex: Agreed, dropped in 0c746fa2139d38481f7917752d5f2b653c5dcc2c

pyproject.toml
```diff
@@ -60,13 +63,11 @@ optional = true
...
 [tool.poetry.group.integration.dependencies]
 pytest = ">=7.2"
-juju = "^3.1.6"
-coverage = { extras = ["toml"], version = ">7.0" }
-pytest-operator = ">0.20"
 # NOTE(boto): 1.36+ incompatible with minio because of MD5 checksum
 boto3 = ">=1.28.0,<1.36"
 tenacity = "^8.2.2"
 spark-k8s-test = "^0.0.2"
+jubilant = "^0.5.0"
```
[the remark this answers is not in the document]
Batalex: Done in 7807b5e315dd79fcc7533e358b745d19071bb285. It's nice, we can reduce logs verbosity and use the new idle helper

PR #78, opened 2025-05-02
[MISC] Implement peer relation with Juju secrets to store password

theoctober19th:
Now Kyuubi will use peer relation (with Juju Secrets) to store password locally (because it needs them when user calls `get-password` action), and the authentication database will store the password hash only.

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -50,3 +50,4 @@ DEPENDENCIES = {
 }
 
 SECRETS_APP: list[str] = []
+PASSWORD_SECRET_SUFFIX = "-password"
--- a/src/managers/auth.py
+++ b/src/managers/auth.py
@@ -92,12 +92,12 @@ class AuthenticationManager(WithLogging):
         status, _ = self.database.execute(query=query, vars=vars)
         return status
 
-    def get_password(self, username: str) -> str:
+    def get_password(self, username: str) -> str | None:
         """Returns the password for the given username."""
         secret_name = self._generate_secret_name(username)
         password = self.context.cluster.relation_data.get(secret_name, None)
         if password is None:
-            raise Exception(f"Could not fetch password for {username}.")
+            self.logger.error(f"Password for {username} not found in peer relation data.")
         return password
 
     def set_password(self, username: str, password: str) -> None:
--- a/src/managers/auth.py
+++ b/src/managers/auth.py
@@ -8,9 +8,8 @@
 import secrets
 import string
 
-from constants import (
-    POSTGRESQL_DEFAULT_DATABASE,
-)
+from constants import PASSWORD_SECRET_SUFFIX, POSTGRESQL_DEFAULT_DATABASE
+from core.context import Context
 from core.domain import DatabaseConnectionInfo
 from managers.database import DatabaseManager
 from utils.logging import WithLogging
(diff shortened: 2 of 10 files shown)


tests/integration/test_provider.py
```diff
@@ -126,19 +124,12 @@ async def test_kyuubi_client_relation_joined(
...
 
     # Assert that a new user had indeed been created
     assert num_users != 0
 
-    logger.info(f"Relation user's username: {kyuubi_username} and password: {kyuubi_password}")
-
-    assert await validate_sql_queries_with_kyuubi(
-        ops_test=ops_test, username=kyuubi_username, password=kyuubi_password
-    )
-
```
Batalex, line 141: **question**: Why don't we test this anymore?

src/managers/auth.py
```diff
@@ -76,35 +109,47 @@ def delete_user(self, username: str) -> bool:
...
+        success, _ = self.database.execute(query=query, vars=vars)
+        if success:
+            secret_name = self._generate_secret_name(username)
+            self.context.cluster.update({secret_name: password})
+        else:
             raise Exception(f"Could not update password of {username}.")
 
     def create_admin_user(self) -> bool:
         """Create a default admin user in the authentication database."""
+        if self.user_exists(self.DEFAULT_ADMIN_USERNAME):
```
Batalex, line 136: **issue**: Hum, if the user already exists, we might have a problem there. Let's say that we remove the relation and then re-relate. The app databag would be empty, so we cannot fetch the admin password with the auth manager `get_password` method.
If the user already exists, we need to update its password in the database so that we can have it in the secret as well

src/managers/auth.py
```diff
@@ -76,35 +109,47 @@ def delete_user(self, username: str) -> bool:
         status, _ = self.database.execute(query=query, vars=vars)
         return status
```
Batalex, line 110: **todo**: Let's also remove the user password from the app databag

Batalex (changes requested):
Overall LGTM, but I have a few concerns regarding the auth flow

src/constants.py
```diff
@@ -50,3 +50,4 @@
 }
 
 SECRETS_APP: list[str] = []
```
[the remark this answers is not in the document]
theoctober19th: I intentionally refrained from adding `admin-password` harcoded, since I think the passwords for other users in addition to `admin` should probably also be supported in the future. The logic with the suffix supports any username and hence I saw no point in using `SECRETS_APP`.

tests/integration/test_provider.py
theoctober19th: We read the password from the `kyuubi_users` table in the authentication database, and connected to Beeline using this password previously.

But now the password won't be in the database in the first place, and hence the password collected from `kyuubi_users` database won't work with Beeline.

Batalex: My point was about testing that client credentials still work as they should. Even if we don't get them from the auth db itself, we should be able to get them from the test charm, right?

deusebio (approved):
The changes look fine. I believe it is still worth asking @delgod a review given this has to do with security, and that he had thoughts on this. The query here matches with what I believe he had in mind, where the checks are done server side using masking function (the crypt and where clause) such that the username/password information never makes it to the client, but I'd like to double check this

tests/integration/test_provider.py
theoctober19th: That's a valid point. I'll address this in the new PR which supersedes this PR here: https://github.com/canonical/kyuubi-k8s-operator/pull/96

src/managers/auth.py
theoctober19th: This will no longer be relevant because we don't have `get-password` now. The changes here are being moved to a new PR: https://github.com/canonical/kyuubi-k8s-operator/pull/96

PR #80, opened 2025-05-09
[DPE-7109] Defer events when data is unavailable and workload is not ready

theoctober19th:
The PR contains the following changes:

* Update all charm libs to latest versions
* Run tasks like preparation of auth_db, metastore_db, etc  by leader unit only
* Defer events when workload is not ready for the events that interact with the workload
* Defer event when database parameters are not available in relation data


src/events/metastore.py
```diff
@@ -55,12 +55,20 @@ def __init__(self, charm: KyuubiCharm, context: Context, workload: KyuubiWorkloa
     @defer_when_not_ready
     def _on_metastore_db_created(self, event: DatabaseCreatedEvent) -> None:
         """Handle event when metastore database is created."""
-        self.logger.info("Metastore database created...")
+        if not (metastore_db := self.context.metastore_db) or metastore_db is None:
```
Batalex, line 58: **nitpick**: We have the same issue in `events/auth.py`, maybe I was going for something else at the time of writing. The second part of the condition is redundant

src/events/metastore.py
```diff
@@ -55,12 +55,20 @@ def __init__(self, charm: KyuubiCharm, context: Context, workload: KyuubiWorkloa
...
+            self.metastore_manager.initialize(schema_version=HIVE_SCHEMA_VERSION)
+            self.logger.info("Metastore database created...")
 
     @compute_status
+    @defer_when_not_ready
     def _on_metastore_db_relation_removed(self, event) -> None:
         """Handle event when metastore database relation is removed."""
-        self.logger.info("Mestastore database relation removed")
         self.kyuubi.update(set_metastore_db_none=True)
+        self.logger.info("Mestastore database relation removed")
```
Batalex, line 74: **typo**: :nerd_face: 
```suggestion
        self.logger.info("Metastore database relation removed")
```

src/events/metastore.py
theoctober19th: Updated in commit 83cd763

Batalex (approved):

welpaolo (approved):
LGTM

PR #84, opened 2025-05-22
[DPE-7397] Force update status of units when auth db and metastore db are created

theoctober19th:


Batalex (commented):
This is something I thought about quite a lot anytime I needed to trigger something on non-leader units. And indeed, it is a simple yet effective solution.
However, I don't like it that much here.

We can find context-aware alternatives to `refresh-token`. For `auth.py`, the admin password will be shared using an app secret. Non-leader units could observe such a secret-changed event to then trigger `.upgrade()` and changed their status accordingly.  
For the metastore, we could add a dedicated field in the peer relation e.g. `{"metastore-schema": "<version>"}` (and reset this field on relation broken).

IMO this is better because now a non-leader unit can infer everything about the current state of the app from (1) what it can access on its own and (2) what it can learn from the databag.
If the leader unit changes, then it is easier to pick up from where the last one stopped, compared to using a refresh token.
We can also be more strict about our statuses now, because having the authentication db info in the authdb relation does not mean that auth is enabled, now we can be more precise:
- no `auth-db` -> Blocked status "Please integrate with auth db"
- `auth-db` but no app secret -> Waiting status "Setting up authentication"

Same with the metastore

On the event handler themselves, we can refactor a little bit and just early return if the unit is not the leader. Non-leader units would overwrite the config files and restart on the peer-relation-changed, with the possibility of having multiple code paths here depending on what actually changed in the databag.

Also, the current implementation of peer-relation-changed only checks if we need a new certificate, so the unit status would indeed change, but the workload is not restarted.

deusebio:
Yes, I would also prefer to rather use actual information. When the leader does an operation, it would write the result of that operation in the peer databag, and this will notify the followers. The refresh token seems a workaround to wake up other units, but would not unlock more advanced logic, for instance the status handling that Alex proposes, which seems quite nice.

deusebio (commented):
Yes, I would also prefer to rather use actual information. When the leader does an operation, it would write the result of that operation in the peer databag, and this will notify the followers. The refresh token seems a workaround to wake up other units, but would not unlock more advanced logic, for instance the status handling that Alex proposes, which seems quite nice.

PR #90, opened 2025-06-05
[DPE-7460] Remove get-jdbc-endpoint action

Batalex:
# Changes

- This PR removes the `get-jdbc-endpoint` action and replace it with an integration with data-integrator in the tests where we previously relied on it
- While doing so, I noticed that we were not updating related clients in case of enabling TLS or changing the service. This is now fixed for LB. I created a ticket for properly enabling nodeport, as we have no support for the moment in the existing client provider logic

TODO after merging: change the quick start guide on discourse

### DIFF
--- a/src/events/provider.py
+++ b/src/events/provider.py
@@ -54,6 +54,44 @@ class KyuubiClientProviderEvents(BaseEventHandler, WithLogging):
             self.database_provides.on.database_requested, self._on_database_requested
         )
 
+    def update_clients_endpoints(self) -> None:
+        """Update related clients.
+
+        Used if expose-external or TLS is changed
+        """
+        if not self.charm.unit.is_leader():
+            return
+
+        service_manager = ServiceManager(
+            namespace=self.charm.model.name,
+            unit_name=self.charm.unit.name,
+            app_name=self.charm.app.name,
+        )
+
+        kyuubi_endpoint = service_manager.get_service_endpoint(
+            expose_external=self.charm.config.expose_external
+        )
+        if kyuubi_endpoint is None:
+            return
+
+        # FIXME: What about nodeports? Same in _on_database_requested
+        jdbc_uri = f"jdbc:hive2://{kyuubi_endpoint.host}:{kyuubi_endpoint.port}/"
+        endpoint = f"{kyuubi_endpoint.host}:{kyuubi_endpoint.port}"
+
+        for client in self.context.client_relations:
+            self.database_provides.set_endpoints(
+                client.id,
+                endpoint,
+            )
+
+            self.database_provides.set_uris(client.id, jdbc_uri)
+            self.database_provides.set_version(client.id, self.workload.kyuubi_version)
+            self.database_provides.set_tls(
+                client.id, "True" if self.context.cluster.tls else "False"
+            )
+            if self.context.cluster.tls:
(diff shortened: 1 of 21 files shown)


welpaolo (approved):
LGTM!

src/events/provider.py
```diff
@@ -54,6 +54,44 @@ def __init__(self, charm: KyuubiCharm, context: Context, workload: KyuubiWorkloa
...
+            app_name=self.charm.app.name,
+        )
+
+        kyuubi_endpoint = service_manager.get_service_endpoint(
+            expose_external=self.charm.config.expose_external
+        )
+        if kyuubi_endpoint is None:
+            return
+
+        # FIXME: What about nodeports? Same in _on_database_requested
```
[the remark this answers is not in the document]
theoctober19th: We initially opted with single endpoint in the past because the defacto client (beeline for us) did not support connecting to a multiple endpoint separated by commas. For example, the client needs to do parsing logic to split up the URLs / endpoints into multiple and choose one of them randomly in order to connect. However, let's spark a discussion on whether we should look back on it and fix it.

tests/integration/conftest.py
```diff
@@ -88,18 +88,25 @@ def charm_versions() -> IntegrationTestsCharms:
         ),
         zookeeper=TestCharm(
             name="zookeeper-k8s",
-            channel="3/edge",
+            channel="3/stable",
             revision=70,
```
theoctober19th, line 92: maybe update the revision to match the `3/stable` channel as well?

tests/integration/conftest.py
```diff
@@ -88,18 +88,25 @@ def charm_versions() -> IntegrationTestsCharms:
...
             base="ubuntu@22.04",
             alias="zookeeper",
         ),
         tls=TestCharm(
             name="self-signed-certificates",
-            channel="edge",
+            channel="latest/stable",
             revision=163,  # FIXME (certs): Unpin the revision once the charm is fixed
```
theoctober19th, line 99: same as above

tests/integration/test_external_access.py
```diff
@@ -79,22 +77,21 @@ def test_nodeport_service(
...
+    )
 
     # Run SQL tests against JDBC endpoint
-    jdbc_endpoint = fetch_jdbc_endpoint(juju)
     assert run_sql_test_against_jdbc_endpoint(
         juju, test_pod=test_pod, jdbc_endpoint=jdbc_endpoint, username=username, password=password
     )
     assert is_entire_cluster_responding_requests(
-        juju, test_pod=test_pod, username=username, password=password
+        juju, test_pod=test_pod, jdbc_endpoint=jdbc_endpoint, username=username, password=password
```
theoctober19th, line 89: in reference to our discussion earlier about returning multivalued endpoints / uris -- in that case here we won't be able to stuff in the endpoint we receive from data-integrator. The client needs to properly parse, split and choose among the multiple endpoints.

theoctober19th (approved):
Solid work! I believe it makes sense to spark discussion on how to go forward with the NodePort service, and whether we need to look back a little bit on the current implementation of returning a single endpoint / URI to the user.

tests/integration/conftest.py
Batalex: Done in 289b0e3b8a3fd88dad6a594c21f2a862fc0b7660!

PR #92, opened 2025-06-11
[DPE-7535] Defer database_requested event if missing auth-db

Batalex:
# Changes

- Address a bug I found while removing the `get-jdbc-endpoint` action in the bundle. The relation with data-integrator was missing the username and password fields if it was created before we had authentication, which is the case in a bundle deployment.


welpaolo (approved):
LGTM

PR #94, opened 2025-06-16
[DPE-7535] Defer database_requested event if user not created

Batalex:
Fix a race condition happening on the bundle CI.


theoctober19th (approved):
LGTM

welpaolo (approved):

PR #98, opened 2025-06-25
[DPE-7589] Add profile config option

welpaolo:
This PR added the profile config option to limit the amount of resources used while testing.

### DIFF
--- a/src/config/spark.py
+++ b/src/config/spark.py
@@ -46,6 +46,12 @@ class SparkConfig(WithLogging):
                     "spark.dynamicAllocation.shuffleTracking.enabled": "true",
                 }
             )
+        if self.charm_config.profile == "testing":
+            conf.update(
+                {
+                    "spark.kubernetes.executor.request.cores": "0.1",
+                }
+            )
         return conf
 
     def _sa_conf(self):
--- a/src/config/spark.py
+++ b/src/config/spark.py
@@ -49,7 +49,8 @@ class SparkConfig(WithLogging):
         if self.charm_config.profile == "testing":
             conf.update(
                 {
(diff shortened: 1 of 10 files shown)


theoctober19th (approved):
The changes look good, thanks for the PR.  I've left some comments, but aren't super critical to block the PR.

src/config/spark.py
```diff
@@ -46,6 +46,12 @@ def _base_conf(self):
                     "spark.dynamicAllocation.shuffleTracking.enabled": "true",
                 }
             )
+        if self.charm_config.profile == "testing":
+            conf.update(
+                {
+                    "spark.kubernetes.executor.request.cores": "0.1",
```
deusebio, line 52: **question, todo** Why only the executor? I believe we generally set the request cores for the driver as well (and I would believe this woulc be important here too since kyuubi spawn a driver as well). It seems that we generally also set these properties to `100m`, see [spark-k8s fixtures](https://github.com/canonical/spark-k8s-bundle/blob/main/python/spark_test/fixtures/service_account.py#L58), [spark-client-snap tests](https://github.com/canonical/spark-client-snap/blob/3.5/edge/tests/integration/ie-tests.sh#L96) and [charmed-spark-rock](https://github.com/canonical/charmed-spark-rock/blob/3.4-22.04/edge/tests/integration/integration-tests.sh#L147).

I would probably keep options consistent unless we have reasons to swap to 0.1, and do this across all testing

deusebio (commented):
The changes look fine, there is only a questions/todo I'd like to address before approving

src/config/spark.py
welpaolo: Done!

deusebio (approved):
Looks good!!!

PR #112, opened 2025-07-17
[DPE-7507] Fix handling of nodeport endpoints

welpaolo:
This PR address the handling of kyuubi endpoints when nodeport is choosen.

### DIFF
--- a/src/managers/service.py
+++ b/src/managers/service.py
@@ -80,7 +80,7 @@ class ServiceManager(WithLogging):
         # Example: kyuubi-k8s-service.my-model.svc.cluster.local
         return f"{self.service_name}.{self.model_service_domain}"
 
-    def _get_node_host(self) -> str:
+    def _get_node_host(self, unit_name: str) -> str:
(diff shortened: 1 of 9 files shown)


src/managers/service.py
```diff
@@ -32,6 +32,9 @@ class Endpoint:
     def __str__(self) -> str:  # noqa: D105
         return f"{self.host}:{self.port}"
 
+    def __hash__(self) -> int:  # noqa: D105
+        return hash((self.host, self.port))
+
```
Batalex, line 37: **thought**: We could perhaps make `Endpoint` "immutable" by adding `frozen=True` which _should_ auto generate a hash implementation. No change required, just thinking out loud

Batalex (approved):

theoctober19th (approved):
Looks good. One nit: might worth renaming `get_service_endpoint` to `get_service_endpoints` since it obviously returns a list now instead of just one endpoint.

PR #115, opened 2025-07-22
[DPE-7484] standardize ci release process

deusebio:


.github/workflows/release.yaml
```diff
@@ -6,18 +6,62 @@ name: Release to Charmhub edge
...
+          BRANCH=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}
+          echo "branch=${BRANCH}" >> $GITHUB_OUTPUT
+          echo "risk=${BRANCH##*\/}" >> $GITHUB_OUTPUT
+          echo "track=${BRANCH%*\/*}" >> $GITHUB_OUTPUT
+        id: branch_metadata
+
+      - name: Extract metadata
+        shell: bash
+        run: |
+          VERSION=(grep "\w*# spark-" metadata.yaml | sed "s/.*spark-\([0-9.]*\).*/\1/g")
```
Batalex, line 32: **suggestion**: Relying on a free-form comment in `metadata.yaml` might not be so robust, especially with a regex. We could inspect the workload with `skopeo`
Here is a one liner to get the same value:
`yq '.resources.kyuubi-image.upstream-source' metadata.yaml | xargs printf -- 'docker://%s' | xargs -0 rockcraft.skopeo inspect --no-creds | yq '.Labels."org.opencontainers.image.version"'`

If too complex, then we should make a note in `metadata.yaml` so that we don't forget to properly update the comment with the correct syntax

deusebio: Yes, I agree with you, and I was not very happy while implementing this. But I didn't want to overcomplicate this, as we need this (and the PR #114 to land by this week) for GA. I'm not sure the command you provide would work in the CI, as it probably requires to have downloaded the image already (beside having docker installed). Also, it would be nice if we could just curl this info from GHCR, but I'm not sure we can, and I would have needed a bit more exploration. 

My suggestion would be to merge this (which I have tested to work on 3.5 branch also) and then work on improving the logic (but this can also be done after GA)

Batalex: Alright, let's move this discussion to after GA

We do indeed need to install rockcraft or skopeo, but we do not need to download the image which is a big plus compared to using the docker CLI to achieve the same result

Batalex (approved):

welpaolo (approved):

PR #125, opened 2025-08-26
[DPE-7941] Add GPU support

Batalex:
This PR adds support for running hardware-accelerated workloads using Apache Kyuubi, following the current state of DA211.

## Changes

- Added the following configuration options
  - gpu-enable
  - gpu-engine-executors-limit
  - gpu-pinned-memory
  - driver-pod-template
  - executor-pod-template
  - executor-cores
  - executor-memory 
- Added a workflow for running a new integration test on an EC2 instance based on #120. We have to move files around and switch users to avoid snap confinement issues. I'm happy to detail why it ended up looking like that.
- The GPU Rock image was missing some jars. I opened https://github.com/canonical/charmed-spark-rock/pull/160 to address this. Once we get a new revision, we need to update the digest here.

## To be done
We settled on having the minimal pod template be part of the Rock image (https://github.com/canonical/charmed-spark-rock/pull/165).
Once it is merged, this PR just needs to:
- Update the OCI digests
- Remove the on-start event handler in `events/kyuubi.py`

Which should be changes trivial enough to not re-request approvals.

### DIFF
--- a/src/events/kyuubi.py
+++ b/src/events/kyuubi.py
@@ -84,7 +84,7 @@ spec:
         """
 
         self.workload.write(
-            gpu_template, f"{self.workload.paths.spark_conf_path}/gpu_executor_template"
+            gpu_template, f"{self.workload.paths.spark_conf_path}/gpu_executor_template.yaml"
         )
 
     @defer_when_not_ready
--- a/src/events/kyuubi.py
+++ b/src/events/kyuubi.py
@@ -46,7 +46,6 @@ class KyuubiEvents(BaseEventHandler, WithLogging):
         self.tls_manager = TLSManager(context=self.context, workload=self.workload)
 
         self.framework.observe(self.charm.on.install, self._on_install)
-        self.framework.observe(self.charm.on.start, self._on_start)
         self.framework.observe(self.charm.on.upgrade_charm, self._on_kyuubi_upgrade)
         self.framework.observe(self.charm.on.kyuubi_pebble_ready, self._on_kyuubi_pebble_ready)
         self.framework.observe(self.charm.on.update_status, self._update_event)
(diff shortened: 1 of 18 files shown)


src/events/kyuubi.py
```diff
@@ -66,6 +67,25 @@ def __init__(self, charm: KyuubiCharm, context: Context, workload: KyuubiWorkloa
     def _on_install(self, _: ops.InstallEvent) -> None:
         """Handle the `on_install` event."""
 
+    @defer_when_not_ready
```
welpaolo: Can we get rid of this now that the image is published ?

welpaolo (approved):
Great work Alex !

src/events/kyuubi.py
Batalex: Removed in 274086bb52968d74413883a6dce71865c5f314a4, as part of the OCI resource update!

tests/integration/conftest.py
```diff
@@ -58,31 +58,31 @@ def charm_versions() -> IntegrationTestsCharms:
     return IntegrationTestsCharms(
         s3=TestCharm(
             name="s3-integrator",
-            channel="edge",
-            revision=41,
+            channel="1/stable",
+            revision=145,
```
deusebio, line 62: **praise** great catch to update these! I'm wondering if we should remove the pinning and test the latest on the stable channels by default. What do you think?

tests/integration/test_gpu.py
```diff
@@ -0,0 +1,85 @@
...
+    assert len(executor_pods) == 1
+
+    exec_pod = executor_pods[0].pod_name
+    logs = get_pod_logs(juju, exec_pod)
+
+    assert (
+        "ExecutorPluginContainer: Initialized executor component for plugin com.nvidia.spark.SQLPlugin"
+        in logs
+    )
+
```
deusebio, line 88: **todo** could we also check that the driver is running on CPU? Would this be a low-hanging fruit?

deusebio (approved):
Great work Alex! The code is extremely clean and clear! I just suggest to also add another assertion (if that is easy enough), but otherwise all good. I don't think I need to re-review.

tests/integration/conftest.py
Batalex: We should probably define a strategy with the team, yes. An alternative could be to test against edge, so that we catch potential regressions as soon as possible. 
I would prefer using stable, though.
Closing this thread, as I am opening a discussion on MM.

tests/integration/test_gpu.py
Batalex: Done in 28cb788!

PR #141, opened 2025-09-12
[BACKPORT] Add GPU support (#125)

Batalex:
Backport #125

### DIFF
--- a/.github/workflows/ci-gpu.yaml
+++ b/.github/workflows/ci-gpu.yaml
@@ -1,0 +1,145 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+name: GPU Tests
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  pull_request:
+  schedule:
+    - cron: "53 0 * * *" # Daily at 00:53 UTC
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+      - name: Install tox
+        run: pipx install tox
+      - name: Run linters
+        run: tox run -e lint
+
+  unit-test:
+    name: Unit tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+      - name: Install tox
+        run: pipx install tox
+      - name: Run tests
+        run: tox run -e unit
+
+  build:
+    strategy:
+      matrix:
+        path:
+          - .
+    name: Build charm | ${{ matrix.path }}
+    uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v32.0.0
+    with:
+      path-to-charm-directory: ${{ matrix.path }}
+
+  start-runner:
+    name: Start self-hosted EC2 runner
+    needs: [build]
+    runs-on: ubuntu-22.04
+    outputs:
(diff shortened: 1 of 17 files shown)


.github/workflows/ci-gpu.yaml
```diff
@@ -0,0 +1,145 @@
...
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_REGION }}
+      - name: Start EC2 runner
+        id: start-ec2-runner
+        uses: machulav/ec2-github-runner@v2.4.1
+        with:
+          mode: start
+          github-token: ${{ secrets.WELPAOLO_GITHUB_TOKEN }}
```
theoctober19th, line 69: question: will we replace this with a canonical Github Token before merging?

config.yaml
```diff
@@ -2,64 +2,153 @@
...
     type: string
     description: The mode in which the service should be exposed externally. Valid values are false, nodeport and loadbalancer.
     default: "false"
-  loadbalancer-extra-annotations:
-    type: string
-    description: Optional extra annotations to be supplied to the load balancer service.
-    default: "{}"
-  enable-dynamic-allocation:
+
+  gpu-enable:
```
theoctober19th, line 53: suggestion -- use the same consistent pattern of `enable-gpu` similar to `enable-dynamic-allocation`.

src/charm.py
```diff
@@ -207,6 +207,23 @@ def _collect_domain_statuses(self) -> list[ops.StatusBase]:  # noqa: C901 - igno
         if not k8s_manager.is_service_account_valid():
             statuses.append(Status.INVALID_SERVICE_ACCOUNT.value)
 
+        if self.config.gpu_enable and not k8s_manager.get_number_of_gpus():
```
theoctober19th, line 210: suggestion -- `is_gpu_enabled()`

welpaolo (approved):
LGTM

src/core/config.py
```diff
@@ -22,18 +22,25 @@
...
-    service_account: str
-    expose_external: ExposeExternal
-    loadbalancer_extra_annotations: str
+    driver_pod_template: str
     enable_dynamic_allocation: bool
+    executor_cores: PositiveInt | None
+    executor_memory: PositiveInt | None
+    executor_pod_template: str
+    expose_external: ExposeExternal
+    gpu_enable: bool
```
theoctober19th, line 31: same as above -- `enable_gpu` is a better name

theoctober19th (approved):
LGTM overall. I've left some minor comments that would be nice to have them addressed.

.github/workflows/ci-gpu.yaml
Batalex: It seems we don't have one as of today. This is the same setup as the one in charmed-spark-rock

config.yaml
Batalex: This was discussed this at length in DA211 where we settled on `gpu-enable` rather than `enable-gpu`. The tradeoff between the existing `enable-*` prefix consistency and the `gpu-*` prefix consistency for everything gpu-related nudged towards the latter form.

src/charm.py
Batalex: We can set up an alias for a field in a pydantic model, though I would ask for your understanding of rejecting this suggestion in this backport PR. The main reason is that even if I would definitely agree with you if it was a classic property, right here it's a pass-through attribute to the charm configuration option. It is less surprising to have this 1-to-1 match here than to introduce additional cognitive complexity with a field alias.

src/core/config.py
Batalex: Based on the previous threads, I'll resolve this one

PR #142, opened 2025-09-16
[DPE-8342] Rework metrics alerts

Batalex:
It turns out that we already had alerts configured. I reworked the existing rules to have something more in line with the rest of DP products and better documented. 

Alerts get properly fired: in the screenshot below, I scaled down a deployment down to 2 units, which triggered the `KyuubiHighAvailibilty` alert.
Then, I messed up with the object storage by changing its credentials. This means that even with the proper Kyuubi credentials, the Kyuubi Engine cannot start. The relevant alert was then fired as well.

image

Using custom alert rules and dashboards is documented in https://github.com/canonical/spark-k8s-bundle/pull/147

### DIFF
--- a/src/prometheus_alert_rules/prometheus.yaml
+++ b/src/prometheus_alert_rules/prometheus.yaml
@@ -1,21 +1,56 @@
-"groups":                                         
-- "name": "kyuubi.alerts"
-  "rules":
-  - "alert": "KyuubiBufferPoolCapacityLow"
-    "annotations":
-      "message": "Buffer pool capacity is low"
-      "summary": "Low buffer pool capacity"
-    "expr": |
-      kyuubi_buffer_pool_direct_capacity < 1000
-    "for": "2m"
-    "labels":
-      "severity": "critical"                                   
-  - "alert": "KyuubiJVMUptime"
-    "annotations":
-      "message": "JVM Uptime"
-      "summary": "JVM Uptime"
-    "expr": |
-      kyuubi_jvm_uptime == 0
-    "for": "1m"
-    "labels":
-      "severity": "critical"                                   
+groups:
+  - name: kyuubi.alerts
+    rules:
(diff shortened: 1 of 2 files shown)


src/prometheus_alert_rules/prometheus.yaml
```diff
@@ -1,21 +1,56 @@
-"groups":                                         
-- "name": "kyuubi.alerts"
-  "rules":
-  - "alert": "KyuubiBufferPoolCapacityLow"
-    "annotations":
-      "message": "Buffer pool capacity is low"
-      "summary": "Low buffer pool capacity"
-    "expr": |
-      kyuubi_buffer_pool_direct_capacity < 1000
```
deusebio, line 9: **question** Is this replaced by the `KyuubiJvmMemoryFillingUp`? Would it make sense to have both? Why was it removed?

Batalex: I could not find anything supporting this specific alert rule online, while keeping heap usage under 80% was pretty much the gold standard of JVM deployments monitoring in every resource I could find.
I can definitely add it back, though. Would you prefer that?

welpaolo (approved):
LGTM

src/prometheus_alert_rules/prometheus.yaml
deusebio: @welpaolo do you have context why we placed this alert back in the days? is this covered by the other alerts we have introduced?

welpaolo: I was not able to track down why this rule was introduced and I was not able to find any documentation to support the presence of rule. My suggestion is to remove it.

deusebio: all right then!

deusebio (approved):

PR #149, opened 2025-10-09
[MISC] Log creation of a user and password update events

theoctober19th:
Add logs for user creation and password events, which are the only user management related events we perform at the charm level.


deusebio (approved):
Looks good! Just please rebase and make sure CI is happy!

Batalex (approved):

PR #150, opened 2025-10-14
[DPE-8542][DPE-8549] Rework health check

Batalex:
This PR introduces the following changes:

- A new HTTP health check for the Kyuubi workload using the REST frontend. The previous version was causing error logs due to not completing the TLS handshake. Fixing the TCP check would have required two different code paths for health checks (i.e., keeping the former one and adding a call to openssl s_client if TLS is enabled), and the REST frontend will be useful once we need it to administer the cluster. Hence, this new health check.
- Adjustments to our CI and tests:
    - Pre-fetch all images in microk8s 
    - Update Kyuubi base revision for refresh tests to the latest stable
    - Wrap the initial populate call in retries. This needs more explanations:

So we often had errors like this one: ` thrift.transport.TTransport.TTransportException: Could not connect to any of [('10.1.30.205', 10009)]` in the refresh tests on the call to `test_populate`. Considering where it would happen in the execution flow, this was either an issue in our previous releases or in the client connecting to Kyuubi.

To rule out (2), I found a way to change our impyla-based client back to pyhive while keeping the TLS capabilities. The issues persisted, but I think this work was valuable enough to keep somewhere (in case we need to use a newer thrift client since impyla pins it to 0.16). You can find a pyhive + TLS with cert verification in this [gist](https://gist.github.com/Batalex/b527412a5f8a7d114cfc64317e2fa9ae).

To figure out why it's happening based on the assumption it's (1), I first tried to update the revision to the latest stable. It did not solve the issue, but I kept the change anyway. SSH-ing into a failing test did not bring any immediately apparent reason as to why the connection would not go through sometimes. Our previous health check might not have been a good one on top of causing errors in logs. Therefore, the call is now wrapped in a retry.

I would say that these changes are a good step in the right direction, as the two latest commits show that all tests passed first try.


deusebio (approved):
All the changes look good and beneficial, so overall +1 from me. 

Also I don't have concerns using the REST endpoint for health check as long as:

1. it is not exposed (since it is not secure and authenticated) 
2. it is a reliable proxy that ALSO the tcp endpoint works 

Regarding (2), the only concern I might have is if it could happen that the REST endpoint works but not the TCP one, but it may be a rather edge case (I would expect the process on the server to error out before). 

But would also implementing the TCP check be hard?

Batalex:
> But would also implementing the TCP check be hard?

No, it would not be hard at all, we would just have a condition as follows:

```py
# src/core/workload/kyuubi.py

    def is_serving_requests(self, is_tls_enabled: bool) -> bool:
        if is_tls_enabled:
            subprocess.check(["bash", "-c", "echo '' | openssl s_client -connect 127.0.0.1:10009"])
        else:
            # previous socket lib solution
```

```diff
# src/charm.py
# where we compute the status

-        if not self.workload.is_serving_requests():
+        if not self.workload.is_serving_requests(self.context.cluster.tls):
            statuses.append(Status.NOT_SERVING_REQUESTS.value)
``` 

I don't really like having those two different code paths for the health check, but it's doable.

welpaolo (approved):
Great and nit work Alex!

PR #152, opened 2025-10-22
[MISC] Pin juju agent to 3.6.9

Batalex:
This PR pins juju agent to 3.6.9 while we figure out what is going on with 3.9.11 and the Integration-hub <-> Kyuubi relation


deusebio (approved):
Yes, I have also seen some instabilities on KF after the release of 3.6.11. Investigating there too

welpaolo (approved):
LGTM

PR #155, opened 2025-10-31
[MISC] Migrate to microceph

Batalex:
Similar to the history-server and the integration hub, this PR migrates our CI to Microceph.
We kept the `fetch_s3_credentials.sh` script so far, but we could make the call to remove it right now.


welpaolo (approved):
LGTM

theoctober19th (approved):
LGTM

PR #164, opened 2026-01-23
[PRA-156] Add ARM support

Batalex:
### **Highlight**: ARM support

This PR introduces ARM support for the charm and updates the OCI resources accordingly.
The test coverage is lacking compared to amd64, since we are missing ARM support from a few charms we need to integrate with.

Refresh tests do not run on arm64 due to not having a previous build for arm64. We now need to assess if we need those or if the extra coverage is not worth doubling the # of tests.

### CI

On the CI side, this PR reorganizes a bit the various integration tests, as the main file was getting a bit long and difficult to navigate in.
While the `charmed-spark-gpu` rock is tested on both amd64 and arm64 with Canonical K8s (because the `nvidia` MicroK8s addon does not support arm64), this simply was not possible here due to (1) missing support for arm64 from the ZooKeeper charm and (2) an issue to investigate regarding Kyuubi + GPU on Canonical K8s. More details in #165. 

This PR completes the migration to microceph and drops the last bits supporting minio.

### Misc.

- Updated matrix link in `metadata.yaml`.


tests/integration/conftest.py
```diff
@@ -57,58 +59,79 @@ def pytest_addoption(parser):
...
             base="ubuntu@24.04",
             alias="self-signed-certificates",
         ),
         data_integrator=TestCharm(
             name="data-integrator",
             channel="latest/stable",
-            revision=161,
-            base="ubuntu@22.04",
+            revision=revisions["data"],
+            base="ubuntu@24.04" if platform == "amd64" else "ubuntu@22.04",
```
Batalex: This one is particularly annoying, this is the only instance where we have to specify a different base based on the platform

.github/workflows/ci.yaml
```diff
@@ -168,89 +203,34 @@ jobs:
...
-          for model in $(juju models --format yaml | yq '.models[] | .name');
-          do
-            echo "Model: $model"
-            if [[ ${model} != *"controller"* ]];then
-              juju debug-log --model $model --replay
-            fi
-          done
+  integration-gpu-arm:
+    name: GPU tests (arm64)
+    if: false # no ARM support for ZK
```
theoctober19th: question: why is this `false`?

.github/workflows/refresh-tests.yaml
```diff
@@ -0,0 +1,107 @@
...
+
+      - name: Setup microceph
+        id: microceph
+        run: |
+          sudo snap install microceph
+          sudo microceph cluster bootstrap
+          sudo microceph disk add loop,1G,3
+          sudo microceph enable rgw
+          sudo microceph.radosgw-admin user create --uid test --display-name test --access-key=foo --secret-key=bar
+          echo -e "S3_SERVER_URL=http://$(ip -4 -j route get 2.2.2.2 | jq -r '.[] | .prefsrc'):80/\nS3_ACCESS_KEY=foo\nS3_SECRET_KEY=bar" > .env
```
theoctober19th: nit: this line looks quite unreadable, perhaps break the URL parsing into separate line?

theoctober19th (commented):
Solid work, Alex! I absolutely love the new restructuring a lot. A couple of questions 👇

.github/workflows/ci.yaml
Batalex: I added this placeholder so that we don't forget to include GPU tests on arm64 once we have arm64 support for ZooKeeper and after we figure out why Kyuubi + GPU does no work on Canonical K8s + gpu-operator installed with the official helm chart.
I'll add a proper `TODO` there and in the integration tests skipped because of the missing ZK charm since it's on the way

Batalex: Documented everything in 3dc5578f7e0b652030153d149070ea0f1435c244!

.github/workflows/refresh-tests.yaml
Batalex: Splitted as part of 3dc5578f7e0b652030153d149070ea0f1435c244

theoctober19th (approved):
Thanks for the additional comments in the code.

tests/integration/conftest.py
Batalex: Nevermind, found the noble build. The revision numbers are hard to find

welpaolo (approved):
Great work Alex!

PR #169, opened 2026-02-19
[PRA-73] Test with latest juju version

Batalex:
## Changes

- Now that we have our compatibility with juju 3.6.13+ back, let's test with the latest available!
- Moved the concierge file for GPU testing in tests resources, as it was one of the feedbacks of the roasting session


welpaolo (approved):
LGTM

theoctober19th (approved):
Solid work, thanks for unpinning the Juju agent version!

PR #180, opened 2026-03-27
[MISC] Fix CI

Batalex:
This PR fixes the CI issues we've been having the past week. 
The issue came from the latest OCI resource for the Integration hub having a breaking change with secrets. While debugging I tried a bunch of different things, hence the pretty hefty diff. The actual fix is bumping the integration hub charm revision.

This PR:
- updates our dependencies and the revisions we use for the charms we need in the integration tests.
- moves the spark8t lib to the integration test dependency group, we don't use it in the actual charm

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -74,7 +74,7 @@ parts:
       # rpds-py (Python package) >=0.19.0 requires rustc >=1.76, which is not available in the
       # Ubuntu 22.04 archive. Install rustc and cargo using rustup instead of the Ubuntu archive
       rustup set profile minimal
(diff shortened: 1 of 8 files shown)


deusebio (approved):
Interesting! We definitely need to keep in mind the issue with the pinning of JUST revisions in tests. Maybe we could re-consider to pin a track, not a revision - in general. We can of course revert to pin revisions when the CI breaks and we are blocked, but in general, maybe it would be best to ONLY use tracks, pin when needed, and when the issue is resolved, resume to use tracks. 

Let's discuss this in our of our dailies

theoctober19th (approved):
LGTM

PR #196, opened 2026-04-03
[PRA-42] Add support for Spark 4.0

theoctober19th:
Add support for Spark 4.0 in Kyuubi K8s Operator

### DIFF
--- a/tests/integration/refresh/test_refresh.py
+++ b/tests/integration/refresh/test_refresh.py
@@ -37,7 +37,7 @@ TABLE_NAME = "inplace_table"
 METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
 
 # spark-3.4.4, release date 01/01/25
-# It should not be necessary to use a 3.5 specific image here.
+# It should not be necessary to use a 4.0 specific image here.
 WORKLOAD_IMAGE_UPGRADE = "ghcr.io/canonical/charmed-spark-kyuubi@sha256:86fc84c8d01da25f756bebbae17395ef9702a8fd855565a4a80ed5d4f8024708"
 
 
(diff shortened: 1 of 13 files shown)


tests/integration/test_metastore.py
```diff
@@ -267,6 +267,12 @@ def test_read_write_with_valid_schema_metastore_again(
     juju: jubilant.Juju, charm_versions: IntegrationTestsCharms
 ) -> None:
     """Test whether previously written data can be read as well as new data can be written."""
+    logger.info(
+        "Waiting for extra 30 seconds as cool-down period for metastore to fully initialize "
+        "after reintegration..."
+    )
+    time.sleep(30)
```
Batalex, line 274: **question**: is not this covered by the `juju.wait` blocking call we have? Should we increase the delay/timeout instead of having this sleep?

Batalex (approved):

tests/integration/test_metastore.py
theoctober19th: In this case, I observed that the unit reporting to `active` status vs the metastore being ready were async. Without this sleep, the tests were a little flaky since previously we only waited for unit / app being active.

tests/integration/refresh/test_refresh.py
```diff
@@ -37,7 +37,7 @@
 METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
 
 # spark-3.4.4, release date 01/01/25
```
welpaolo, line 39: Can you update this as well ? And the image below ?

welpaolo (approved):
LGTM

tests/integration/refresh/test_refresh.py
theoctober19th: Looking at the comment alongside the image, I think this entry need not be the same as the spar-4.0.0 used for workload (as was the same for 3.5/edge). I'm gonna leave this as is for now, as refresh tests are disabled now anway, and if necessary we can change when we reenable refresh tests.

Issue #197, opened 2026-04-06
Kyuubi does not work with S3 TLS


theoctober19th:
The Kyuubi charm is unable to work with S3 TLS when the S3 relation comes via the integration hub.

This is because the Integration Hub upon relation with S3 creates some Kubernetes Secret (for the truststore) and then adds Spark configuration to mount that secret into the driver and executor pods. This works fine as long as Spark is initialized in `cluster` mode. But in `client` mode (which Kyuubi uses by default), the Kyuubi pod itself initializes the Spark Driver JVM -- the Kyuubi pod is not mounted with the Secret that contains the truststore. As a result, the Kyuubi JVM fallbacks to its default truststore, and thus fails to initialize the Kyuubi engine at all.

## Steps to reproduce

1. Deploy MicroCeph with custom TLS setup (for reference, see [here](https://github.com/canonical/object-storage-integrator/blob/main/s3/tests/integration/conftest.py#L126) )
2. Deploy S3 Integrator and configure it with the S3 credentials and TLS chain
3. Deploy Spark Integration Hub, and relate it with S3 Integrator
4. Deploy data-integrator  and get Kyuubi credentials
5. Connect to Kyuubi JDBC using beeline.
6. Observe that the Kyuubi Engine fails initialization.

## Expected behavior
The Kyuubi connection works just fine.

## Actual behavior

The following logs are seen in the Beeline CLI:
```bash
bikalpa@bikalpacodes:~/essentials$ /opt/spark-3.5.6-bin-hadoop3/bin/beeline -u "jdbc:hive2://10.1.139.107:10009/" -n relation_id_11 -p R76FGR9b8g3kLSo7
Connecting to jdbc:hive2://10.1.139.107:10009/
Error: org.apache.kyuubi.KyuubiSQLException: org.apache.kyuubi.KyuubiSQLException: 	Suppressed: software.amazon.awssdk.core.exception.SdkClientException: Request attempt 1 failure: Unable to execute HTTP request: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
	Suppressed: software.amazon.awssdk.core.exception.SdkClientException: Request attempt 2 failure: Unable to execute HTTP request: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
[...]

PR #200, opened 2026-04-15
[PRA-243] Enable the usage of S3 provider with custom CA with Kyuubi

theoctober19th:
The PR aims to fix the bug #197, by mounting the secret received from integration hub via resource-manifest manually by the Kyuubi charm.

The PR also enables the usage of S3 TLS in the integration tests, such that such bugs will be caught by the CI in the future.


src/managers/integration_hub.py
```diff
@@ -0,0 +1,96 @@
...
+
+        truststore_secret = next(
+            (
+                m
+                for m in manifests
+                if isinstance(m, dict)
+                and m.get("kind") == "Secret"
+                and isinstance(m.get("metadata"), dict)
+                and isinstance(m["metadata"].get("name"), str)
+                and m["metadata"]["name"].startswith(TRUSTSTORE_SECRET_PREFIX)
```
Batalex: **thought**: No change required. I think we could have simplified this big condition by using the pattern-matching syntax to describe what the content of the secret we look for should look like

```python
truststore_secret = None
for m in manifests:
    match m:
        case {
            "kind": "Secret", 
            "metadata": {"name": str(name)}
        } if name.startswith(TRUSTSTORE_SECRET_PREFIX):
            truststore_secret = m
            break
```

src/managers/kyuubi.py
```diff
@@ -54,6 +58,33 @@ def _compare_and_update_file(self, content: str, file_path: str) -> bool:
...
+
+        In client deploy mode, Kyuubi itself runs the Spark driver process, so the truststore
+        must exist in the Kyuubi pod filesystem.
+        """
+        if not service_account_info:
+            return
+
+        # Always clean previously managed truststores so stale files do not survive
+        # when integration hub removes or rotates truststore secrets.
+        for path in self.workload.list("/"):
```
Batalex, line 71: **nitpitck**: IMO it would be cleaner to have the truststore in `/etc/kyuubi` rather than the root, but I can live with this

src/managers/kyuubi.py
```diff
@@ -109,6 +141,10 @@ def update(
                     ).contents,
                     self.workload.paths.kyuubi_properties,
                 ),
+                self._compare_and_update_file(
+                    KyuubiEnvironConfig(service_account_info=service_account_info).contents,
+                    self.workload.paths.kyuubi_env,
+                ),
```
Batalex, line 145: **praise**: Love this pattern you came up with so much 🫶

tests/integration/conftest.py
```diff
@@ -144,12 +144,16 @@ def s3_bucket_and_creds(request: pytest.FixtureRequest) -> Iterable[S3Info]:
     access_key = os.environ["S3_ACCESS_KEY"]
     secret_key = os.environ["S3_SECRET_KEY"]
     endpoint_url = os.environ["S3_SERVER_URL"]
+    region = os.environ.get("S3_REGION", "us-east-1")
```
Batalex, line 148: **question**: Which change required the addition of the region? Nothing against it, I'm just curious.
Subsequent question: since we test with microceph, do we have anything against `"default"`?

tests/integration/helpers.py
```diff
@@ -500,6 +501,12 @@ def deploy_minimal_kyuubi_setup(
     secret_key = s3_bucket_and_creds["secret_key"]
     bucket_name = s3_bucket_and_creds["bucket"]
     path = s3_bucket_and_creds["path"]
+    region = s3_bucket_and_creds["region"]
+    ca_bundle_path = s3_bucket_and_creds["ca_bundle_path"]
+    tls_ca_chain = ""
+    if ca_bundle_path:
+        with open(ca_bundle_path, "r") as f:
+            tls_ca_chain = base64.b64encode(f.read().encode()).decode()
```
Batalex: **suggestion**: No change required. I noticed that you reused this bit of code in the test_charm, how about adding a new field to the dictionary returned by the fixture instead?

Batalex (approved):
Incredibly well-done fix, congrats!
I like everything: the new `env.py`, the new hub manager, the CI setup.

I left a few comments, but I'd be happy if we proceeded with the merge as-is

src/managers/kyuubi.py
theoctober19th: The only reason I kept it at `/` is because it's where the Integration Hub mounts this secret at. As a result, if we do the same in Kyuubi pod, we would not need to parse and change the Spark property sent to us via Integration Hub.

The idea is, leave the Spark property we receive from Integration Hub as is, while Kyuubi does the patch to fix the file being at the exact location Integration Hub assumed when it created the Spark properties.

tests/integration/conftest.py
theoctober19th: One thing I felt with the existing setup was that it was not straightforward to run tests locally using the S3 provider of your choice. The `default` vs `us-east-1` is just whether we consider Ceph vs AWS S3 default behavior, and I chose AWS S3 here. The recent boto3 versions expect region to be a non-empty value.

src/managers/kyuubi.py
Batalex: Yes, it makes perfect sense

tests/integration/helpers.py
theoctober19th: Nice suggestion, updated this in 15e5642.

src/managers/integration_hub.py
theoctober19th: Thanks for the suggestion. I usually don't prefer match-case unless we really need it, as I feel the syntax a little confusing and unnecessarily complex. I'm gonna stick with the simple if condition.

src/common/workload/__init__.py
```diff
@@ -53,10 +53,40 @@ def write(self, content: str, path: str, mode: str = "w") -> None:
         """
         ...
 
+    @abstractmethod
+    def write_bytes(self, content: bytes, path: str) -> None:
```
deusebio: **suggestion** could we embed this into the `write` method? Maybe we can just allow for content to be either `str` or `bytes`, or managed this with the `mode` allowing a `wb` 

Seems to me a bit too verbose to create a method only for this use case that can be embedded in the existing method.

src/common/workload/__init__.py
```diff
@@ -53,10 +53,40 @@ def write(self, content: str, path: str, mode: str = "w") -> None:
...
+        """Deletes a file or directory from the workload.
+
+        Args:
+            path: the full filepath to delete
+            recursive: whether to delete directories recursively
+        """
+        ...
+
+    @abstractmethod
+    def list(self, path: str) -> list[str]:
```
deusebio, line 67: **praise** this is a good method! yes, that's a shame we were missing it to manage ops operations. Eventually we could probably use the `ops.Path` abstraction, but this should be handled in a dedicated ticket

src/managers/integration_hub.py
```diff
@@ -0,0 +1,96 @@
...
+import re
+
+import yaml
+
+from constants import TRUSTSTORE_SECRET_PREFIX
+from core.domain import IntegrationHubTrustStore, SparkServiceAccountInfo
+from utils.logging import WithLogging
+
+
+class IntegrationHubManager(WithLogging):
```
deusebio, line 14: **praise** nice refactoring. This was indeed a good idea to encapsulate this into its own manager, since it was really not as much K8s specific, more integration hub specific. Well done!

src/managers/integration_hub.py
```diff
@@ -0,0 +1,96 @@
...
+        """Return whether Azure object storage backend has been configured."""
+        pattern = r"spark\.hadoop\.fs\.azure\.account\.key\..*\.dfs\.core\.windows\.net$"
+        return any(re.match(pattern, prop) for prop in self.spark_properties)
+
+    def is_executor_pod_template_configured(self) -> bool:
+        """Return whether executor pod template has been configured."""
+        pattern = r"spark\.kubernetes\.executor\.podTemplateFile$"
+        return any(re.match(pattern, prop) for prop in self.spark_properties)
+
+    def get_hub_truststore(self) -> IntegrationHubTrustStore | None:
```
deusebio: **suggestion** this seems to be a parsing logic, and as such, it would possibly be better placed in `core`. I'm thinking that this can just be done also when creating the `service_account_info` object. I would rather focus here not on parsing data structures/validation, but business logic, e.g. the pattern matching rules above are quite appropriate. Parsing dataclasses, less so

It is not super critical, but also the refactoring would probably be simple as well and make the code a bit more clean IMHO

deusebio (commented):
The code base looks fine. I have really liked the extension of the workload abstraction as well as encapsulating the IntegrationHub logic into its own manager. 

However I have a couple of points, that also not-super blocking, also seems appropriate to have a more structured and maintanable code base. Please have a look, and check whether they are sensible

src/managers/integration_hub.py
theoctober19th: Fixed in d706136.

src/common/workload/__init__.py
theoctober19th: Fixed in d706136.

welpaolo (approved):
Great work Bikalpa! I like the final solution

deusebio (approved):
Looks very nice! Well done!

PR #201, opened 2026-04-17
[PRA-76] Renovate configuration to update OCI resources

Batalex:
This PR is similar to the other charms. 
However, we have a slightly more complex setup, since we also want to track and update the resources used to run the spark job.

Here is an example in my own fork of what a renovate PR would look like: https://github.com/Batalex/kyuubi-k8s-operator/pull/1

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -3,11 +3,7 @@
   "extends": [
     "github>canonical/data-platform//renovate_presets/charm.json5",
   ],
-  "reviewers": [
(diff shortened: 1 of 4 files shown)


theoctober19th (approved):
LGTM. Similar comment as https://github.com/canonical/spark-integration-hub-k8s-operator/pull/195#discussion_r3099007990

welpaolo (approved):
LGTM

PR #206, opened 2026-04-17
[BACKPORT] [PRA-243] Enable the usage of S3 provider with custom CA with Kyuubi

theoctober19th:
Backport of #200 for 4.0/edge


Batalex (approved):
LGTM, once we confirm the proper resolution of the permission issues on 3.5/edge we should bring the improvements here!

welpaolo (approved):

PR #212, opened 2026-04-22
[PRA-264] Fix TIOBE workflow

Batalex:
This PR fixes the broken TIOBE workflow that had gone unnoticed for a while.

The root cause was simple: we (and by "we", I mean "I") introduced a while ago a test using `skopeo` to check the compatibility between the various OCI resources involved.

Instead of installing the tool in the runner, I deselected this specific test since it does not contribute to the testing coverage and makes external network calls. I don't have the strongest of opinions here, so I'm happy to proceed the other way around and install `skopeo` on the runner.

You can check in the action tab that we now have a green run from the workflow (and even check the dashboard with a new quality measurement from today). Since I removed the commit in this PR running the workflow on pull request, the thing is a bit hidden.

### DIFF
--- a/.github/workflows/tiobe_scan.yaml
+++ b/.github/workflows/tiobe_scan.yaml
@@ -47,7 +47,7 @@ jobs:
           echo PATH="$PATH" >> "$GITHUB_ENV"
 
       - name: Run tox tests to create coverage.xml
-        run: tox run -e unit
+        run: tox run -e unit -- -m "not skopeo"
 
       - name: Move results to necessary folder for TICS
         run: |
--- a/.github/workflows/tiobe_scan.yaml
+++ b/.github/workflows/tiobe_scan.yaml
@@ -1,7 +1,6 @@
 # Copyright 2025 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-
 name: TICS run self-hosted test (github-action)
 
 on:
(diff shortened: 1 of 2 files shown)


theoctober19th (commented):
One way we can still keep the test and also not have to install `skopeo` in the runner machine is to install `rockcraft` snap in the runner, and use `skopeo: str` to provide `rockcraft.skopeo` value. I don't have strong opinion on keeping / removing the test from `tox -e unit` as long as it is possible to invoke this test somewhere separately in the CI (such that we don't leave gaps updating image with wrong version).

Batalex:
> One way we can still keep the test and also not have to install `skopeo` in the runner machine is to install `rockcraft` snap in the runner, and use `skopeo: str` to provide `rockcraft.skopeo` value. I don't have strong opinion on keeping / removing the test from `tox -e unit` as long as it is possible to invoke this test somewhere separately in the CI (such that we don't leave gaps updating image with wrong version).

Oh yes, we definitely *could* install the tool in the runner, whether it be through the rockcraft snap or the vanilla tool itself. We could also just skip the test if `skopeo` is not found in the `PATH`, which is something I would prefer not to do.

welpaolo (approved):

theoctober19th (approved):

PR #233, opened 2026-05-05
[MISC] Enable renovate on track 4, fix oci updates

Batalex:
This PR fixes the OCI update so that #221 does not happen again. See https://github.com/Batalex/kyuubi-k8s-operator/pull/1 for an example where the 3.4 branch only receive 3.4 updates.

In addition, it also enables renovate on track 4.


theoctober19th (approved):

welpaolo (approved):
LGTM

PR #239, opened 2026-06-01
[PRA-63] Run integration tests using spread

Batalex:
This PR is similar to what we did in https://github.com/canonical/spark-history-server-k8s-operator/pull/171.

It is worth noting that:
- In the refresh tests, TLS for microceph was always disabled. I tried to enable it for consistency with the rest of the integration test, but it appears that rev 112 (first stable release) was incompatible at the time of release with TLS-RGW.
- I took this opportunity to extract the "basic" integration tests into their own workflow. `ci.yaml` runs the linter, unit tests and build jobs, then passes the artifacts to the integration tests, refresh tests and GPU tests on EC2


.github/workflows/refresh-tests.yaml
```diff
@@ -5,106 +5,167 @@ name: Inplace refresh tests
...
+    name: ${{ matrix.job.name }}
+    needs:
+      - collect-refresh-test
+    runs-on: ${{ matrix.job.runner }}
     timeout-minutes: 120
     steps:
-      - name: (GitHub hosted) Free up disk space
-        shell: bash
+      - name: Free up disk space
+        timeout-minutes: 10
```
theoctober19th, line 93: I'm rather surprised this operation would take that long.. I remember this being extended to something around 2 mins a while back when a lot of CI runs timed out on the 1 min limit. No changes needed, though.

spread/integration-refresh/task.yaml
```diff
@@ -0,0 +1,58 @@
...
+  # Pre-pull old revision OCI images needed as the starting point for refresh tests
+  OLD_IMAGES=(
+    "ghcr.io/canonical/charmed-spark-kyuubi@sha256:e2029c6976fc5b9ee2865eced632ca42ce554039d9832af20dfa3e63113e00f7" # rev 112 charm OCI resource
+    "ghcr.io/canonical/charmed-spark@sha256:f1f944369108c0b0112212fb0242f3c314dfad362926c234857029f13c5de2c0" # rev 112 job OCI resource
+  )
+  for IMAGE in "${OLD_IMAGES[@]}"; do
+    microk8s ctr images pull "$IMAGE"
+  done
+
+  # The old charm revision deployed by these tests does not handle the self-signed
```
theoctober19th, line 34: +1 for the nice comment with explanation.

tests/integration/setup/concierge.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+juju:
+  channel: 3.6/stable
+  model-defaults:
+    logging-config: <root>=INFO; unit=DEBUG
+
+providers:
+  microk8s:
+    enable: true
+    channel: 1.32-strict/stable
```
theoctober19th, line 12: **thoughts**: should we perhaps test with more recent stable channels, like 1.34?

tox.ini
```diff
@@ -111,7 +110,7 @@ commands =
 
 
 
-[testenv:integration-{charm,trust,ha,upgrade,external-access,tls,dynamic-allocation,observability,iceberg,metastore,auth,provider,refresh,gpu}]
+[testenv:integration-{charm,trust,ha,external-access,tls,dynamic-allocation,observability,iceberg,metastore,auth,provider,refresh,gpu}]
```
theoctober19th, line 113: I see the `integration-upgrade` has been removed, is this intentional?

theoctober19th (commented):
Thank you so much for the PR, this was a tech debt we wanted to clear from long ago.

.github/workflows/refresh-tests.yaml
Batalex: Yeah I just copied and pasted it from a different source after removing it rather from the previous commit because I had it in a different tab, hence the changes

tests/integration/setup/concierge.yaml
Batalex: I believe the next step would be to migrate to the k8s snap instead (possibly with 1.34+, sure), for a future PR

tox.ini
Batalex: Yes it is intentional, this tox target was not doing anything. It was replaced by the refresh tests, but the target was never removed here

theoctober19th (approved):
LGTM. Thanks a lot!

welpaolo (approved):

PR #244, opened 2026-06-16
chore: upgrading postgresql to track 16

deusebio:


Batalex (approved):
The future is here 🚀

theoctober19th (approved):
The changes look good. The only thing I'd also request to do is have this tested against the bundle and the UATs to be 100% sure that the upgrade works without any surprises.

PR #281, opened 2026-07-10
[PRA-340] feat: Support for LDAP authentication Kyuubi charm

theoctober19th:
This pull request introduces support for LDAP authentication to the charm, alongside several related improvements and refactorings. The main changes include new configuration options, interface requirements, and code updates to handle both JDBC and LDAP authentication modes, as well as improved TLS environment handling. Additionally, new integration test scaffolding is added, and some code is refactored for clarity and extensibility.

**LDAP Authentication Support and Related Features:**

* Added `ldap-credentials` as a required relation in `metadata.yaml` and defined `LDAP_RELATION_NAME` in `constants.py`, enabling the charm to connect to an LDAP service for authentication. [[1]](diffhunk://#diff-d186204ea7abb85bbadd9d94ab29d5bcefcc43d6a0ed06015e98a71b2f099423R51-R54) [[2]](diffhunk://#diff-42d0f00c4780305311eae71fecca993a0523587cb4156e039cabd0d1ad8dbbb8R22-R23)
* Introduced a new `ldap-search-filter` config option in `config.yaml` to allow customization of the LDAP search filter used for user authentication.
* Updated `core/context.py`, `core/config.py`, and `config/kyuubi.py` to handle LDAP authentication configuration, including validation logic to block the charm if both JDBC and LDAP are configured or if neither is set. [[1]](diffhunk://#diff-47d39b3220fbaf4926feaf4d7579dc0c271c8998d108fb2a9dee1c7a5ceebf37R14-R28) [[2]](diffhunk://#diff-47d39b3220fbaf4926feaf4d7579dc0c271c8998d108fb2a9dee1c7a5ceebf37R43) [[3]](diffhunk://#diff-2e8a8cf531fb260742fa3ab78a64041219ec1a5763e05d07434ff221eceeb3efR42) [[4]](diffhunk://#diff-0eaafaa0674d3b9d7862c6c17c26b0906b0e6ca45d3e40bd39a8d11089aa8010L10-R10) [[5]](diffhunk://#diff-0eaafaa0674d3b9d7862c6c17c26b0906b0e6ca45d3e40bd39a8d11089aa8010L22-R30) [[6]](diffhunk://#diff-0eaafaa0674d3b9d7862c6c17c26b0906b0e6ca45d3e40bd39a8d11089aa8010R80-R101) [[7]](diffhunk://#diff-0eaafaa0674d3b9d7862c6c17c26b0906b0e6ca45d3e40bd39a8d11089aa8010R113-R135)
* Enhanced domain status checks in `charm.py` to reflect the new authentication logic, including new status messages for missing or conflicting authentication relations.

**TLS and Certificate Handling Improvements:**

* Added `receive-ca-cert` as an optional relation for certificate transfer and updated related constants and dependencies. [[1]](diffhunk://#diff-d186204ea7abb85bbadd9d94ab29d5bcefcc43d6a0ed06015e98a71b2f099423R75-R77) [[2]](diffhunk://#diff-42d0f00c4780305311eae71fecca993a0523587cb4156e039cabd0d1ad8dbbb8R22-R23) [[3]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R19) [[4]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R120-R121)
* Improved TLS environment variable handling in `config/env.py`, allowing the charm to correctly set up truststore information for backend TLS connections. [[1]](diffhunk://#diff-825bf726d829ec62f46c5b314c51cde00a8a2aa7867090b045b57f5e6ea193b3L8-R8) [[2]](diffhunk://#diff-825bf726d829ec62f46c5b314c51cde00a8a2aa7867090b045b57f5e6ea193b3R18-R23) [[3]](diffhunk://#diff-825bf726d829ec62f46c5b314c51cde00a8a2aa7867090b045b57f5e6ea193b3R41-R61)

**Workload Abstractions and Temporary File Support:**

* Added an abstract `temporary_file` context manager to the workload interface and implemented it for Kubernetes-based workloads, enabling safe creation and cleanup of temporary files in the workload container. [[1]](diffhunk://#diff-2ef4de114fe397496e74de33f7200ee7e69599440edf9bccb5a9bac7cacc9c17R8) [[2]](diffhunk://#diff-2ef4de114fe397496e74de33f7200ee7e69599440edf9bccb5a9bac7cacc9c17R84-R101) [[3]](diffhunk://#diff-844c9c1039f8863a5724537e5b9cab2f4e54b8982a6324b960e0532a68a62308R8-R11) [[4]](diffhunk://#diff-844c9c1039f8863a5724537e5b9cab2f4e54b8982a6324b960e0532a68a62308R93-R114)

**Integration Test and CI Workflow Updates:**

* Added new integration test scaffolding for LDAP authentication (`spread/integration-ldap-auth/task.yaml`) and JDBC authentication (`spread/integration-jdbc-auth/task.yaml`), and updated the CI workflow to account for arm64 limitations and new test tasks. [[1]](diffhunk://#diff-0d9c26f1c33f7c4a9ece70556bbdcfcb5a6b5856c501f723837410785244facbR1-R8) [[2]](diffhunk://#diff-292b6309206afeda5ba95583b088af1cdeddb6a1f2adbcff962e82ab7bc55514R1-R8) [[3]](diffhunk://#diff-b6766eb8febc0c51651250cd0cdfb44c4f0d3256470d88e62bf82fd46aa73ae0R60-R66)
* Removed obsolete or duplicate integration test tasks.

**Refactoring and Minor Improvements:**

* Refactored event handler imports and initialization to clarify authentication event handling. [[1]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cL37-R37) [[2]](diffhunk://#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cL74-R80)
* Updated status checks and method calls in `charm.py` for clarity and to match new interfaces.

These changes collectively add robust support for LDAP authentication, improve TLS handling, and lay groundwork for expanded integration testing.


src/config/kyuubi.py
```diff
@@ -75,6 +77,28 @@ def _base_conf(self) -> dict[str, str]:
...
+                "Kyuubi configurations are not generated, the charm will go to blocked state."
+            )
+            return {}
+        elif not (self.db_info or self.ldap):
+            self.logger.warning(
+                "Neither JDBC nor LDAP authentication is configured. "
+                "Kyuubi configurations are not generated, the charm will go to blocked state."
+            )
+            return {}
+        if self.db_info:
```
Batalex: **nitpick**: We could use one less condition path (the final `else` is not possible) + we could have continued the same chain of `elif`s here

metadata.yaml
```diff
@@ -48,6 +48,10 @@ requires:
     interface: postgresql_client
     limit: 1
     optional: false
+  ldap-credentials:
+    interface: ldap
+    limit: 1
+    optional: false
```
Batalex: **question**: should both of them be optional since it's one or the other?

src/core/workload/kyuubi.py
```diff
@@ -110,23 +109,32 @@ def generate_password(self) -> str:
...
+
+        This performs an unauthenticated request against the Kyuubi REST
+        frontend. Since the charm no longer holds valid end-user credentials,
+        we cannot authenticate the request. Instead, we rely on the fact that
+        Kyuubi routes every REST call through its authentication filter: when
+        the server is healthy it answers an unauthenticated request with an
+        HTTP response (``401 Unauthorized`` when authentication is enabled),
+        whereas a server that is not serving requests raises a transport-level
+        error. Any HTTP response therefore means the frontend is up and
+        processing requests.
```
Batalex, line 123: **thought**: oh, that's a shame :/ I wish more products did not gatekeep the health endpoint behind authentication

src/events/auth/ldap.py
```diff
@@ -0,0 +1,59 @@
...
+)
+
+from core.context import Context
+from core.workload.kyuubi import KyuubiWorkload
+from events.base import BaseEventHandler, defer_when_not_ready
+from managers.kyuubi import KyuubiManager
+from utils.logging import WithLogging
+
+if TYPE_CHECKING:
+    from charm import KyuubiCharm
```
Batalex, line 24: **todo**: If you use this pattern, then you probably need `from __future__ import annotations`, otherwise the interpreter will just crash at runtime because it cannot find `KyuubiCharm`

edit: Well, I have NO idea why the tests worked then

src/core/workload/__init__.py
```diff
@@ -37,25 +37,34 @@ def server_key(self) -> str:
...
     @property
     def keystore(self) -> str:
-        """The Java Truststore containing service private-key and signed certificates."""
+        """The Java Keystore containing Kyuubi server's private-key and signed certificates."""
         return f"{self.conf_path}/keystore.p12"
 
+    @property
+    def transferred_certificates(self) -> str:
+        """The certificates transferred to Kyuubi via relations."""
+        return f"{self.conf_path}/transferred-certificates.pem"
```
Batalex: **suggestion**: maybe we can remove this one, seems unused. It confused me because I did not understand why we had those two very similar elements

src/core/domain.py
```diff
@@ -42,7 +42,15 @@ class Status(Enum):
     WAITING_PEBBLE = MaintenanceStatus("Waiting for Pebble")
     MISSING_OBJECT_STORAGE_BACKEND = BlockedStatus("Missing Object Storage backend")
     MISSING_INTEGRATION_HUB = BlockedStatus("Missing integration hub relation")
-    MISSING_AUTH_DB = BlockedStatus("Missing authentication database relation")
+    MISSING_AUTH_RELATION = BlockedStatus(
+        "Missing authentication relation. Please integrate to one of `ldap` or `postgresql_client` interface."
```
Batalex: **suggestion**: To give more actionable feedback, we should refer to the integration, not the interface. At the very least for `auth-db` since we have two integrations relying on the `postgresql_client` interface, but if we do it for one we might as well do it for the other and refer to `ldap-credentials`

src/managers/tls.py
```diff
@@ -116,65 +120,167 @@ def set_private_key(self) -> None:
...
+        bundle_bytes = self.workload.read(
+            self.workload.paths.transferred_certificate_file(relation_id=relation_id)
+        ).encode()
+        certificates = x509.load_pem_x509_certificates(bundle_bytes)
+        return certificates
+
+    def generate_alias_for_certificate(self, certificate: x509.Certificate) -> str:
+        """Generates an alias for the given certificate based on its SHA256 fingerprint."""
+        fingerprint = certificate.fingerprint(hashes.SHA256()).hex()[:16]
+        return f"transferred-cert-{fingerprint}"
```
Batalex: **question**: have you thought of using the relation id as the alias suffix? Since we use it for the file on the disk, this would make it consistent with the content of the truststore, should we manually inspect it

tests/integration/helpers.py
```diff
@@ -558,27 +686,11 @@ def deploy_minimal_kyuubi_setup(
...
-            charm_versions.auth_db.app,
-        ),
-        delay=15,
-        timeout=1000,
-    )
+    if auth_mode == "jdbc":
+        setup_jdbc_authentication(juju=juju, charm_versions=charm_versions)
+    elif auth_mode == "ldap":
+        setup_ldap_authentication(juju=juju, charm_versions=charm_versions)
+        apply_sample_users_ldif(juju=juju, charm_versions=charm_versions)
```
Batalex, line 693: **praise**: Cool and needed abstraction!

Batalex (changes requested):
Very cool test setup, as usual with your work :)

I left a few comments, nothing too major

src/events/auth/ldap.py
theoctober19th: I've added the import line in 55a4932

src/config/kyuubi.py
theoctober19th: You're right. Fixed this in db285ea.

src/core/workload/kyuubi.py
theoctober19th: Yeah I agree. This check here is the best I could find, given the constraints.

src/core/workload/__init__.py
theoctober19th: You're right. I've removed this in f8b2e33

src/managers/tls.py
theoctober19th: I did consider both of the options here. The only reason I preferred the fingerprint of the certificate to using relation ID here was that I found the fingerprint <> certificate more closely correlated and coupled than relation-id <> certificate.

But the other way around also has some pros and cons. The pro being better visibility should we need to inspect, and also support for duplicate certificates (if two relations use the same CA cert) and thus when one of the relations is detached, the certificate is not completely removed from truststore (since another active relation has same CA cert).

Now that you mentioned it and I think about it, I think using relation ID is a better approach compared to using fingerprints, however we also need to be sure that the relation IDs are always the same even after juju refreshes. (or else we will leave some certificates dangling with expired relation ids).

theoctober19th: Oh wait, I remember now why I used fingerprint and not relation ID.

So when I initially implemented this, I wanted to use the certificate_transfer interface lib v1, for which each relation would pass a **list of certificates** to Kyuubi. So for a relation ID A, there can be multiple certificates, and keytool is not able to import them in a batch, so we need separate aliases for each one of them. Using only relation ID was not enough to disambiguate between different certificates in the same relation, so I had to choose something that's even more unique **per certificate**, and it was the fingerprint of the certificate.

However, later I could not use v1 interface because there seems to be some limitation in lib compatibility (see here: ), and settled for v0 for now (where we are importing only CA certificate in this implementation at the moment). I believe the middle ground of using both relation ID and fingerprint would be a good idea here, so as to also ease manual debugging if needed.

theoctober19th: For now, I udpated this logic to use both Relation ID and Fingerprint in commit 91adef3, but I'll leave this thread open to hear from other reviewers.

src/core/domain.py
theoctober19th: Thanks, fixed this in 540d198.

metadata.yaml
theoctober19th: I'm not sure if whether we should mark both as optional=true, or both as optional=false, since even if both of them are optional, one of them is required. It's a little tricky to map out this to boolean.

theoctober19th: I checked [the docs](https://canonical.com/juju/docs/charmcraft/4.3/reference/files/metadata-yaml-file/), and it seems that this key is just informational, and does nothing more than that. I believe it is sensible here to mention both of these interfaces as optional, and let the charm status message tell the users when none of them are integrated.

src/common/workload/k8s.py
```diff
@@ -87,3 +90,25 @@ def exec(
         except ExecError as e:
             logger.error(str(e.stderr))
             raise e
+
+    @override
+    @contextmanager
+    def temporary_file(self, content: str | bytes = "", mode: str = "w") -> Iterator[str]:
```
deusebio, line 96: **praise** I like this abstraction and it feels a useful feature/method. Well done!

Batalex (approved):

src/managers/tls.py
deusebio: yes, using Relation ID and Fingerprint makes sense to me such that it is easy to be debugged. +1

src/events/auth/__init__.py
deusebio, line 1: **question** are we generally using this pattern? Not a bit point, but if this is only for using 

```
from managers.auth import JDBCAuthenticationManager
```

We could also just use

```
from managers.auth.jdbc import JDBCAuthenticationManager
```

and avoid one file. I don't think we really need to be specific about which API is public and to be used in the context of the charm code

src/events/auth/jdbc.py
```diff
@@ -16,19 +16,19 @@
...
 
 
-class AuthenticationEvents(BaseEventHandler, WithLogging):
-    """Class implementing PostgreSQL metastore event hooks."""
+class JDBCAuthenticationEvents(BaseEventHandler, WithLogging):
+    """Class implementing JDBC authentication event hooks."""
 
     def __init__(self, charm: KyuubiCharm, context: Context, workload: KyuubiWorkload) -> None:
-        super().__init__(charm, "s3")
+        super().__init__(charm, "jdbc")
```
deusebio, line 31: **question** what is this string for? is this really important? I don't remember why we set if to certain values, and I suppose it is mostly a placeholder since we also had kind of bug here

src/events/tls.py
```diff
@@ -64,25 +73,40 @@ def __init__(self, charm: KyuubiCharm, context: Context, workload: KyuubiWorkloa
             private_key=private_key,
             refresh_events=[self.refresh_tls_certificates_event],
         )
+        self.certificates_transfer = CertificateTransferRequires(
```
deusebio: **suggestion** I don't really have super strong opinions, but I would have kept this relation into a separate EventHandler class, such that TLS and CertificateTrasfer can be used and plugged independently. It does not seem that the hook are currently using methods of the "TLSEvents" that need to be shared. 

Let's see if others agree or disagree, but ultimately, up to you, @theoctober19th . As I said, it is not a big point.

src/managers/tls.py
```diff
@@ -116,65 +120,169 @@ def set_private_key(self) -> None:
...
             return
 
         self.workload.write(
-            content=self.context.unit_server.certificate, path=self.workload.paths.certificate
+            content=self.context.unit_server.kyuubi_server_certificate,
+            path=self.workload.paths.kyuubi_server_certificate,
         )
 
-    def set_truststore(self) -> None:
+    def set_transferred_certificates(self, relation_id: int) -> None:
```
deusebio: **minor issue** it is a bit of niptick, but ideally, managers should be independent from charm logic, such that they can also be used/tested outside of charms. Having a `relation-id` argument creates a binding to charm logic. I'm just thinking that we could rather write the logic here to handle inputs with `alias` and `x509.Certificate`, and the parsing of the certificate using the relation is actually done in the handlers, not in the managers.

src/events/auth/ldap.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class LDAPAuthenticationEvents(BaseEventHandler, WithLogging):
```
deusebio, line 27: **major** I'm not sure where this is used. I tried to look for the usage of this class and I found none. Are we wiring up the hooks for LDAP?

src/charm.py
```diff
@@ -71,13 +71,13 @@ def __init__(self, *args) -> None:
...
         # Context
-        self.context = Context(model=self.model, config=self.config)
+        self.context = Context(charm=self, config=self.config)
 
         # Event handlers
         self.kyuubi_events = KyuubiEvents(self, self.context, self.workload)
         self.hub_events = SparkIntegrationHubEvents(self, self.context, self.workload)
         self.metastore_events = MetastoreEvents(self, self.context, self.workload)
-        self.auth_events = AuthenticationEvents(self, self.context, self.workload)
+        self.auth_events = JDBCAuthenticationEvents(self, self.context, self.workload)
```
deusebio: **question** I would have expected to also have the EventHandlers for the LDAP relations. This is connected to one comment above about the usage of "LDAPAuthenticationEvents"

tests/integration/test_ldap_auth.py
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/env python3
```
deusebio, line 1: **praise** very good test, very readable and covering both happy and unhappy paths!

deusebio (changes requested):
I have a few minor comments and a major one, that needs to be addressed: I believe the LDAP event handler is not wired up correctly. It is interesting that we didn't spot this with the tests, probably because we only rely on integration tests and update status or config changes may have masked this, from an integration perspective. However, unittests would have possibly flagged this. Indeed in this PR, we have not added many unittests despite the UX changes, and I would definitely add unittests on relation changed events to make sure that - using scenario - adding/removing LDAP relations has the expected effect.

src/events/auth/ldap.py
Batalex: @theoctober19th  To help you with understanding what is going on, the charm lib for the LDAP requirer is instantiated in context.py, hence why the integration tests work

theoctober19th: You're right, this was not explicitly wired, but somehow worked anyway because of the context.py file. I've added an explicit registering of these event handlers in the charm.py file in commit e9b0f34.

src/charm.py
theoctober19th: You're right, I have fixed this in e9b0f34.

src/events/auth/jdbc.py
theoctober19th: This second param is the `key` which the `ops.Object` parent class takes from the constructor. This is the unique ID of any object we instantiate in `ops`. More information in the following docstring taken from `ops`:

```
class Object:
    """Initialize an Object as a new leaf in :class:`Framework`, identified by `key`.

    Args:
        parent: parent node in the tree.
        key: unique identifier for this object.

    Every object belongs to exactly one framework.

    Every object has a parent, which might be a framework.

    We track a "path to object," which is the path to the parent, plus the object's unique
    identifier. Event handlers use this identity to track the destination of their events, and the
    Framework uses this id to track persisted state between event executions.

    The Framework should raise an error if it ever detects that two objects with the same id have
    been created.
```

theoctober19th: I believe it would work as long as we have "unique" keys, but it is also very sensible to make those unique keys be self-describing about what this object will eventually be used for.

src/events/auth/__init__.py
theoctober19th: I think since we move the files into a new package under the `auth` directory, we would still need the `__init__.py` file such that Python knows the `auth` directory is a package. As far as the content of this file is, however, is opinionated, and I preferred to keep the import path clean (and would keep it unless there's strong opinion to keep the file empty).

src/managers/tls.py
theoctober19th: Thanks for the suggestion, applied this in e468c18.

src/events/tls.py
theoctober19th: Thanks for the suggestion. I've splitted the file into two in commit e1c052f.

theoctober19th:
@deusebio I have made the following changes in the PR:

1. Refactor the TLS manager to decouple it from the `relation_id`, and have those logic done in event handler.
2. Addition of unit tests using Scenario in the context of LDAP
3. Split the `tls` event handlers into a distinct `certificate_transfer` event handler.
4. Other minor fixes.

These address all of your concerns in the previous review. Can you please go through another round of review?

welpaolo (approved):
LGTM

src/events/auth/__init__.py
deusebio: yes yes, you are right, the `__init__.py` would be there but empty. I would in general use this pattern at the top level of the package to only expose publicly certain classes. But again, it is not a big point. I just found it a bit strange in this context. 

Resolving now!

src/events/auth/jdbc.py
deusebio: All rihgt! thanks for the explanation!

deusebio (approved):
Looks good! Thanks for addressing the comments!

